PR #3, opened 2016-12-17
Improve the BIST testbench and documentation

mithro:


enjoy-digital:
thanks, I'll do so small changes.

PR #12, opened 2018-05-03
Add auto-precharge support

JohnSully:
When the next row to be accessed is different we can embed the precharge command into the read by toggling A10.  This work does that and saves us some command bus traffic.

Its not as efficient area-wise as it could be.  If we could peek into the next cmd_buffer value then we could eliminate the FIFO and row_last registers I added.

Timing was validated via simulation with a Micron Bus Functional model.

### DIFF
--- a/litedram/core/bankmachine.py
+++ b/litedram/core/bankmachine.py
@@ -68,19 +68,6 @@ class BankMachine(Module):
             req.lock.eq(cmd_buffer.source.valid),
         ]
         
-        #Set row change buffer
-        self.comb += [
-            rowchg_buffer.sink.differentRow.eq(slicer.row(req.adr) != row_last),
-            rowchg_buffer.sink.valid.eq(cmd_buffer.source.valid & cmd_buffer.sink.valid & cmd_buffer.sink.ready),
-            rowchg_buffer.source.ready.eq(cmd_buffer.source.ready),
-            auto_precharge.eq(rowchg_buffer.source.differentRow & rowchg_buffer.source.valid),
-        ]
-        self.sync += [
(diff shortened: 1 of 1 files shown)


JohnSully:
This represents about 10% faster memory accesses when page conflicts are frequent.  I wouldn't expect it to have any benefit for sequential memory access patterns.

enjoy-digital:
Thanks! I'll review and test very soon.

enjoy-digital:
Thanks, i merging and will try to improve area if needed.

enjoy-digital:
@JohnSully: can you check that this (https://github.com/enjoy-digital/litedram/commit/8ba7fcab236a311f3481f32232154df191f34378) is equivalent to your commit?

JohnSully:
I believe it is equivalent.

enjoy-digital:
Ok thanks.

enjoy-digital:
I just reverted the changes because this was introducing a regression. I'll need to do more simulations before re-integrating that feature.

JohnSully:
Can you let me know what the issue was?  It will help me not repeat similar mistakes in the future.

enjoy-digital:
I don't know yet what is wrong, just that there is a regression. I'll investigate to see what was going on but needed to revert to avoid breaking others projects.

JohnSully:
This issue? https://github.com/enjoy-digital/litedram/issues/14

JohnSully:
Potential issues from code review:
We shouldn't be checking the precharge timer in the AUTOPRECHARGE state.  We would need to either delay the read/write or avoid auto precharge entirely if timing constraints aren't met.  I think the whole AUTOPRECHARGE state was a bug.

PR #19, opened 2018-07-31
Fix timing issues (tRRD, tCCD, and tFAW)

JohnSully:
Issues:
tCCD was per bank while the timing constraint is per rank.  I moved this to the multiplexer
tRRD was not implemented at all.  It was usually unnecessary with a 1:4 controller.
tFAW was per bank while the timing constraint is per rank.  As with tCCD it is moved to the multiplexer

### DIFF
--- a/litedram/common.py
+++ b/litedram/common.py
@@ -31,7 +31,7 @@ class GeomSettings:
 
 
 class TimingSettings:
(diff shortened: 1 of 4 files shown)


enjoy-digital:
Thanks! merged. (I'm assuming you simulated it with the DDR3 model)

PR #24, opened 2018-08-10
Auto precharge

JohnSully:
This work re-enabled auto-precharge.  To truly shine it needs to be combined with out of order support (and so this PR is dependent upon the Out of Order PR).  I've reworked the timing which should hopefully fix the problems we were having before.

Given our previous problems with this change and my lack of hardware to test with this should be treated with extra scrutiny.


enjoy-digital:
Thanks, i will test it on hardware before merging.

PR #27, opened 2018-08-20
board tuning parameters added

bunnie:


bunnie:
the tuning parameters here match the expected parameters in sdram_init.py

note the defaults set may not match what you were originally using, these are the ones that work on the point-to-point board i've taped out, but if you want to be safe you can set the defaults to what has been tested on your boards before...

enjoy-digital:
Thanks! merged manually with some changes.

The default parameters seems to be the default for point to point:
https://github.com/enjoy-digital/litedram/blob/master/litedram/sdram_init.py#L211

If you need specific electrical parameters for your board, you can now use:
`self.ddrphy.settings.add_electrical_settings(rtt_nom, rtt_wr, ron)`

PR #36, opened 2018-09-08
Fix failing timing

JohnSully:
Some timings are off by 1 core cycle.


enjoy-digital:
Thanks, have you seen this while simulating with a model?

Issue #40, opened 2018-09-13
1:2 Sequential Access Throughput is Lower Than Expected


JohnSully:
The 1:2 Phy only gets 64% bus efficency with a sequential read pattern.  The 1:4 is capable of 97%.  

My theory is the arbitrator in the bank machine mux is to blame.

enjoy-digital:
Thanks. Are you going to have a look or do you want i look at that? (Not sure i'll be able to look at it in the next days). I'm planning to add unit-test of the internal modules of the cores to help figuring out this kind of things.

JohnSully:
I’ll take a look

enjoy-digital:
Fixed with https://github.com/enjoy-digital/litedram/commit/06c8c2afcf1e3c6cff1f7f2f38e6d9b352220055 that is indeed reverting to old code.

PR #42, opened 2018-09-22
We wait an extra cycle for no reason

JohnSully:
I know I broke this and insisted it was right... To be fair there were a lot of other timing errors back then so perhaps it was confusion with one of those.  Anyways the extra cycle in tCCD is what capped 1:2 @ 66% peak performance.  This affects other timings but they don't have the same magnitude of effect.

Issue #40


litedram/core/multiplexer.py
```diff
@@ -136,11 +136,11 @@ def __init__(self, txxd):
         # # #
 
         if txxd is not None:
-            count = Signal(max=max(txxd+1, 2))
+            count = Signal(max=max(txxd, 2))
```
enjoy-digital: If we want to be able to load txxd value in the signal, len must be txxd +1

litedram/core/multiplexer.py
```diff
@@ -136,11 +136,11 @@ def __init__(self, txxd):
...
             self.sync += \
                 If(valid,
                     count.eq(txxd),
-                    If((txxd - 1) == 0,
+                    If((txxd) == 0,
```
enjoy-digital: For me it's doing the opposite, it wait an extra cycle. (load a > value), probably related to the other point.

We need to add unit-tests for these small modules, i'll try to look at that.

litedram/core/multiplexer.py
JohnSully: You are correct, but then the fix stops working so I think the bug is more complex.  Getting closer to a solution.

litedram/core/multiplexer.py
JohnSully: It was supposed to change the line above... it appeared to work because of the first bug.  I'm testing a proper fix now.

litedram/core/multiplexer.py
JohnSully: This is now correct because we set the count to txxd-1

enjoy-digital:
So it seems we are back to the original code?
:https://github.com/enjoy-digital/litedram/commit/efd7a4789021c020cf0ca0534f90364ee4b7d867

PR #43, opened 2018-09-22
Bank valid/ready refactor

JohnSully:
This was originally written as a fix for the 1:2 efficiency bug but it was not the cause.

However I think this refactor is necessary to do smarter reordering in the future.  Right now when cmd.ready is false a bank machine will not assert the valid signal and therefore the multiplexer has no information on which bank to choose next.  Making the valid signal independent of ready is also a cleaner design.

This change does improve one scenario: When a command comes in while we're waiting for tCCD in a different bank than the current (and all other banks are idle).

NOTE: This is a pull request to the new staging branch.  Additional testing is needed before up-streaming to master but I want to code review it.  It simulates correctly.

### DIFF
--- a/litedram/core/bankmachine.py
+++ b/litedram/core/bankmachine.py
@@ -31,7 +31,6 @@ class BankMachine(Module):
         self.req = req = Record(cmd_layout(aw))
         self.refresh_req = Signal()
         self.refresh_gnt = Signal()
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks, it seems fine and cleaner solution. Let's merge and do more testing on this branch.

PR #44, opened 2018-09-23
This adds support for tRC timing parameters

JohnSully:
I'm getting these timing failures on random sequences that are only across 2 banks.  More rarely with wider patterns.


enjoy-digital:
Thanks, merged.

PR #45, opened 2018-09-23
Implement tRAS

JohnSully:


JohnSully:
This is actually too pessimistic for auto precharge as it doesn't account for the delay of performing the read or write.

JohnSully:
Fixed.  Key quotes from the spec on auto precharge:

> Auto precharge ensures that the precharge is initiated at the earliest valid stage within a burst. This earliest valid stage is determined as if an explicit PRECHARGE command was issued at the earliest possible time without violating tRAS (MIN)

AKA we don't need to have any delays on asserting auto_precharge.  We do need to account for tRAS in the AUTOPRECHARGE stage where we wait for when the command would be executed.

PR #46, opened 2018-09-25
Update the write-to-precharge timings so it works with 1:2

JohnSully:
This updates the formula so it doesn't use hardcoded numbers.  I believe it correctly accounts for the total end-to-end write time, however the datasheet mentions a timing parameter AL that I don't see us use anywhere.  The latency between the command and start of transmission of the write data is supposed to be CWL + AL.

Tested with the 1:2 and 1:4 Phys.

### DIFF
--- a/litedram/core/bankmachine.py
+++ b/litedram/core/bankmachine.py
@@ -1,3 +1,4 @@
+import math
 from migen import *
 from migen.genlib.misc import WaitTimer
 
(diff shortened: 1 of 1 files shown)


JohnSully:
Looking at how we initialize MR1, Additive Latency is always 0 and not supported.  So this formula should be correct unless we start using that.

enjoy-digital:
Thanks for looking at this! We are indeed not using Additive Latency so this is fine. (I'll just add a comment here in case we want to use AL in the future)

PR #54, opened 2018-10-02
Add a latency parameter so wdata can be registered

JohnSully:
It's really hard to meet timing in the 1:2 controller.  With the reordering controller its not possible to register the wdata input in all cases as its dependent on the wbank parameter.  This optional feature allows the wbank and ready signals to be asserted N cycles early to allow registering of the data bus after MUXing the bank.

N must be less than CWL, and will usually just be 1.  The setting is per master.

### DIFF
--- a/litedram/common.py
+++ b/litedram/common.py
@@ -120,7 +120,7 @@ def rdata_description(data_width, with_bank):
 
 
 class LiteDRAMNativePort:
-    def __init__(self, mode, address_width, data_width, clock_domain="sys", id=0, with_bank=False):
+    def __init__(self, mode, address_width, data_width, clock_domain="sys", id=0, with_bank=False, write_latency=0):
         self.mode = mode
         self.address_width = address_width
         self.data_width = data_width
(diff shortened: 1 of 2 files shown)


JohnSully:
This is also necessary to use Block RAM to implement the reordering buffer since its synchronous.

enjoy-digital:
Thanks. I think it's better to keep that in the bank_reordering branch for now and we'll integrate that when merging the bank_reordering branch.

JohnSully:
As part of simplifying the bank reordering PR, reopening this.

JohnSully:
Yes there are ports with differing settings.  In my case I have a Wishbone port for debugging which doesn't need this.  

I see it as being a function of which adapter you put over the native port.

JohnSully:
Also depending on the architecture of the port adapter you may want different latencies. 

E.g. if I want a huge reordering buffer on one with block ram that ideally has 2 cycles when fully pipelined. And a port with a small or no reordering buffer for a lesser used part with 0 or 1 cycle latency. 

I wouldn’t be able to mix them if write latency were global.

enjoy-digital:
Sorry, i initially merged it, but i then add to revert some code since one of the merge i did yesterday introduced a regression (seems related to the trtwcon). I'm in the process of stabilizing things without reordering before creating a release, so i'll merge the changes/ideas later.

JohnSully:
Sounds good.

Issue #30, opened 2018-08-22
Add Reordering support


enjoy-digital:
Ports can now expose banks to user and can allow reordering accesses to the memory.

To implement reordering, we could create a module that would work on two native ports:
- one used by the user with all accesses in order and not exposing the banks.
- one used internally with reordered accesses and exposing the banks.

````
user_port = LiteDRAMNativePort(..., with_reordering=False)
internal_port = LiteDRAMNativePort(..., with_reordering=True)

class LiteDRAMReordering(Module)
    def __init__(self, user_port, internal_port):
        [...]
````

We could implement this scenario:

For writes:
- add bank cmd buffers (only store row/col).
- each bank cmd buffers maintain a in /out count.
- add data ram (min depth = nbanks * depth of cmd buffers)
- redirect write cmd to the proper bank cmd buffer.
- write cmd is accepted if the proper bank buffer is not full.
- when bank cmd buffer accepts the cmd, accept the data and store it at location
bank << log2(buffers' depth) + bank in index.
- when bank cmd buffer outputs the cmd, retrieve the data in ram at location
bank << log2(buffers' depth) + bank out count and put it in data queue that will
be presented to the crossbar.

For reads:
- maintain a global cmd_in and data_out count.
- add bank cmd buffers (only store row/col/cmd_in count).
- add data ram (min depth = nbanks * depth of cmd buffers).
- redirect read cmd to the proper bank cmd buffer.
- read cmd is accepted if the proper bank buffer is not full and if read data corresponding to the same cmd_in count value has been presented to the user (should be cmd_in count + 1 != data_out count).
- when a bank accepts a cmd, put cmd_in outputed by the cmd buffer value in a queue, use this queue to know where to store the next returned data in the data ram.
- use a flip bit in data ram to indicate that data has been updated (flip this bit each time a location is used).
- read data at data_out count, if bit has flipped, present the data and increment data_out count to read next location.

JohnSully:
Work has started in the reordering branch.  I'm focusing on optimizing read/write grouping right now which doesn't require an API change.

JohnSully:
At low speeds (~800MT/s) row open/close is well hidden with good dispersion among banks so we don't need to re-order based upon activate/precharge commands.  The primary bottleneck at this speed is read/write bus transitions.  

[...]

PR #55, opened 2018-10-08
Bank reordering

JohnSully:
This is in bank reordering + 1:2 timing fixes.  I've been testing the 1:2 Phy heavily, **I haven't tested 1:4.  This will need to be done prior to merging**.  The 1:4 Phy has different behavior in the bank machine so its quite possible there are issues there.  The changes are quite extensive so I'm issuing the PR early to start discussions.

Analysis shows that the bulk of the delay in random access at speeds achievable by 7 series parts is R/W bus turnarounds.  So this only reorders reads and writes to minimize turnarounds.  I'm seeing 3x improvement in throughput on the random Read+Update pattern. (originally 20%, now 60% bus efficiency).

This reordering does not change the ordering guarantees from the perspective of the client.  We ensure the reads/writes to the same address are issued in sequence and so everyone can benefit.

### DIFF
--- a/litedram/core/bankmachine.py
+++ b/litedram/core/bankmachine.py
@@ -100,8 +100,15 @@ class BankMachine(Module):
             )
 
         # Chose the correct source for the command
-        self.sync += [
+        sync_or_comb = None
+        if settings.timing.tCCD > 1:
+            # This inserts a pipeline bubble of one cycle and so cannot be used when a command is required every cycle
+            sync_or_comb = self.sync
+        else:
+            sync_or_comb = self.comb
+        sync_or_comb += [
             If((self.want_writes & cmd_bufferWrite.source.valid) | ~cmd_bufferRead.source.valid,
+                # Note: when sync valid is asserted one extra cycle (OK because tCCD prevents issuing a duplicate command)
                 cmd_buffer.valid.eq(cmd_bufferWrite.source.valid),
                 cmd_buffer.we.eq(1),
                 cmd_buffer.addr.eq(cmd_bufferWrite.source.addr),
--- a/litedram/core/bankmachine.py
+++ b/litedram/core/bankmachine.py
@@ -33,6 +33,8 @@ class BankMachine(Module):
         self.refresh_req = Signal()
         self.refresh_gnt = Signal()
         self.want_writes = Signal()
+        self.has_writes = Signal()
+        self.has_reads = Signal()
         a = settings.geom.addressbits
         ba = settings.geom.bankbits + log2_int(nranks)
         self.cmd = cmd = stream.Endpoint(cmd_request_rw_layout(a, ba))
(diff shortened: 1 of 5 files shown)


litedram/frontend/crossbar.py
```diff
@@ -163,16 +172,11 @@ def do_finalize(self):
...
-        for i in range(self.write_latency-1):
-            new_master_wbank = Signal(max=self.nbanks)
-            self.sync += new_master_wbank.eq(wbank)
-            wbank = new_master_wbank
 
         for master, master_ready in zip(self.masters, master_readys):
             self.comb += master.cmd.ready.eq(master_ready)
-        for master, master_wdata_ready in zip(self.masters, master_wdata_readys):
-            self.comb += master.wdata.ready.eq(master_wdata_ready)
+#        for master, master_wdata_ready in zip(self.masters, master_wdata_readys):
```
JohnSully: Delete commented code.

litedram/core/bankmachine.py
```diff
@@ -45,73 +49,118 @@ def __init__(self, n, aw, address_align, nranks, settings):
...
             )
 
+        # Chose the correct source for the command
+        sync_or_comb = None
+        if settings.timing.tCCD > 1:
+            # This inserts a pipeline bubble of one cycle and so cannot be used when a command is required every cycle
+            sync_or_comb = self.sync
+        else:
+            sync_or_comb = self.comb
+        sync_or_comb += [
```
JohnSully: Its worth discussing why I did this instead of just using a Migen Buffer.  The buffer's valid signal has a combinatorial dependency on it's ready which has a high logic depth.   Inserting a bubble makes timing significantly easier.

litedram/core/bankmachine.py
```diff
@@ -45,73 +49,118 @@ def __init__(self, n, aw, address_align, nranks, settings):
...
+            If((self.want_writes & cmd_bufferWrite.source.valid) | ~cmd_bufferRead.source.valid,
+                # Note: when sync valid is asserted one extra cycle (OK because tCCD prevents issuing a duplicate command)
```
JohnSully: It's even more subtle than this.  If cmd.valid were deasserted in the bubble cycle then the arbitrator in the multiplixer would move on.  This would require an extra cycle for recovery limiting sequential throughput to 2/3.

enjoy-digital:
Thanks, i'll try to have a closer look soon.

JohnSully:
As unrelated pieces are integrated into master only the reordering parts will remain.  Its a little tricky since some timing fixes rely on the refactor done for reordering.

JohnSully:
Thanks  is_activate is a bug I was trying to find!

JohnSully:
That’s another regression from when I rolled back my banckmachine changes. We definitely need to respect that timing parameter. https://github.com/enjoy-digital/litedram/commit/6275f6cc59b480934407622afad88f0d1e0e0a1d

enjoy-digital:
We can probably close this now, we will not be able to merge it but it could be used for inspiration when doing https://github.com/enjoy-digital/litedram/issues/30.

PR #62, opened 2018-11-06
modules: Add AS4C32M16 32Mx16 SDRAM

daveshah1:
WIP, some reliability issues when testing with ECP5 Versa hat


enjoy-digital:
Thanks, merging now since the issues are probably not related to the module definition.

PR #77, opened 2019-03-02
ecp5ddrphy: Shift read position forwards to fix higher frequencies

daveshah1:
Works reliably for me at 50MHz, 75MHz and 100MHz


enjoy-digital:
Thanks, i also wanted to try that.

PR #79, opened 2019-04-29
sdram_init: use "unsigned long" for address values

gsomlo:
Enable support for future 64-bit architectures: "unsigned long" matches the native pointer size of each architecture (32 vs. 64 bits, respectively), as opposed to "unsigned int" which is always 32 bits, and therefore won't be able to store 64-bit addresses.


enjoy-digital:
Thanks, merged

PR #81, opened 2019-05-15
RFC (don't pull yet): Allowing direct access to CSR (I/O) registers

gsomlo:
I'd like to generate a LiteDRAM controller (or, better yet, LiteX "virtual motherboard") exposing an AXI port for the CPU's cached-RAM access, and a "raw" CSR port for I/O. (As an aside, this would fit really well the RocketChip, which routes cached-RAM access over a separate AXI port from the one it uses for MMIO, so it wouldn't make a lot of sense to have both those AXI ports be masters on the same shared interconnect, but I digress :)

I'd like to take advantage of LiteX commit ef1c7784 which introduced the option to expose direct access to the CSR bus, and have modified the litedram_gen script as shown in the accompanying patch.

The one problem I'm having is that I think line 250 of the patch is probably wrong and doesn't actually connect me to the master interface of the CSR bus, but rather generates unconnected wires in the resulting verilog.

Any help connecting to the real CSR bus much appreciated!

### DIFF
--- a/examples/litedram_gen.py
+++ b/examples/litedram_gen.py
@@ -71,6 +71,16 @@ def get_dram_ios(core_config):
         ),
     ]
 
+def get_csr_ios(aw, dw):
+    return [
+        ("csr_port", 0,
+            Subsignal("adr", Pins(aw)),
+            Subsignal("we", Pins(1)),
+            Subsignal("dat_w", Pins(dw)),
+            Subsignal("dat_r", Pins(dw))
+        ),
+    ]
+
 def get_native_user_port_ios(_id, aw, dw):
     return [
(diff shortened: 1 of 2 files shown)


gsomlo:
Nevermind, I figured it out. If we set cpu_type to none, there will be no wishbone_masters added to SoCCore, and SoCCore's do_finalize method won't actually create and connect the CSRs to the bus [in this block](https://github.com/enjoy-digital/litex/blob/1ea22d49b7d0b0b200c9599836458070c31a5d4b/litex/soc/integration/soc_core.py#L483). Closing, as I need to re-think my approach and maybe resubmit something different, later...

PR #83, opened 2019-05-27
modules/ddr3: add IS43TR16128B_125K

ambrop72:
This is the chip that is actually on my Arty A7 100T (there is no mention of this chip in the Arty reference, which claims it is MT41K128M16JT-125).

I just added the speed class that is on my board, adding others with the intended design might be an issue because I think some of the "technology timings" actually depend on the speed class.

### DIFF
--- a/litedram/modules.py
+++ b/litedram/modules.py
@@ -310,6 +310,19 @@ class K4B2G1646F(SDRAMModule):
     speedgrade_timings["default"] = speedgrade_timings["1600"]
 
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, it's merged. I just renamed it to IS43TR16128B and added the 1600 speedgrade to be consistent with the others definitions. BTW, i also reviewed tRFC: https://github.com/enjoy-digital/litedram/commit/bc88cfa6f79d870690777da8097d1030f2086f50, thanks for the review.

PR #86, opened 2019-07-05
dma: expose reservation level in the reader

sergachev:
I propose to expose the rsv_level of LiteDRAMDMAReader as an attribute.
This change does not break anything and allows for example to start burst read operations only when the FIFO has enough data:

dma_reader = LiteDRAMDMAReader(...)
read_trigger.eq(dma_reader.rsv_level >= video_line_width)


enjoy-digital:
Thanks for the patch and use case explanation. That's merged with a minor simplification: https://github.com/enjoy-digital/litedram/commit/fc4175108092db7afb5bf6a874a57bdede8d803a.

PR #90, opened 2019-08-16
Fusesoc

olofk:
This series adds support for generating parameterized litedram cores on the fly when using FuseSoC. It has been tested with a modified version of the SweRVolf SoC.

I have tried to keep it clean to not affect current users of the example script, but the most controversial change is probably the addition of generate.py in the main litedram dir


enjoy-digital:
Thanks! I was already thinking changing the file format if the config files to JSON or YAML, since it would also be useful for fusesoc, we'll just support a format that could be used by the generator and Fusesoc. The generator should indeed be part of litedram package to ease reuse.

I first need to think a bit before merging, but this is a good direction.

enjoy-digital:
I just switched default config file format to YAML and made the generator accessible when LiteDRAM is installed, you can now just do "litedram_gen config.yml" Since FuseSoC could also be useful to generate specific configuration for others cores, i think it could be better to have a specific repository for FuseSoC generator/configuration files and just created https://github.com/litex-hub/litex-fusesoc. What do you think about regrouping FuseSoC files here?, we could have for example a litedram.core / litedram_gen.py for now and in the future create LiteEth, LiteSATA, etc... generators/configurations. If that's fine for you, can you create a pull request in this repository? If you still think it's better to have FuseSoC files directly in LiteX core, i'm open to discuss that.

olofk:
I agree that a separate repo is better. That way FuseSoC users only need to add that repo as a FuseSoC library. Will submit a PR soon

enjoy-digital:
@olofk: i'm doing some cleanup on LiteDRAM issues/pull request and am curious to know if you end up with something with the FuseSoC integration of LiteDRAM generators? LitePCIe, LiteEth now also have generators, so that could be an interesting approach. If you give me enough information or a skeleton, i could probably try to put something in place in https://github.com/litex-hub/litex-fusesoc.

olofk:
I know what to do but I haven't had time yet :/ I'll try to make something basic that you can look at

enjoy-digital:
Thanks no problem, i understand. If you have something, even basic, feel free to share and i'll try to create something in https://github.com/litex-hub/litex-fusesoc. I'm closing this PR since we'll integrate that differently.

PR #91, opened 2019-08-31
modules: Add support for Micron MT47H32M16 DDR2 RAM

sd-fritze:
Hi,
this adds support for the Micron MT47H32M16 family of DDR2 RAMs.
They are (except for being half the size and having half the bank count) identical to the MT47H64M16 already supported, so they could also be written with less duplication.


enjoy-digital:
Thanks, merged.

PR #92, opened 2019-10-23
frontend/[axi,wishbone]: add data_width match assertion, add base_address to LiteDRAMWishbone2Native

gsomlo:
Rocket's 'mem_axi' and LiteDRAM's native ports both happen to have the same (64-bit) data width, (EDIT: true only on the nexys4ddr -- on the versa, liteDRAM is 128bit!) and therefore can be connected together using LiteDRAMAXI2Native() (see https://github.com/enjoy-digital/litex/pull/284).

However, we should ensure that axi and native ports have the same data width, in case we ever end up trying to connect interfaces with non-matching data width!

### DIFF
--- a/litedram/frontend/axi.py
+++ b/litedram/frontend/axi.py
@@ -32,6 +32,7 @@ class LiteDRAMAXIPort(AXIInterface):
 class LiteDRAMAXI2NativeW(Module):
     def __init__(self, axi, port, buffer_depth, base_address):
         assert axi.address_width >= log2_int(base_address)
+        assert axi.data_width == port.data_width
         self.cmd_request = Signal()
         self.cmd_grant = Signal()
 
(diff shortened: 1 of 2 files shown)


gsomlo:
Not sure about this anymore, e.g. what happens if the axi data_width is less than the native litedram port data_width? We're probably writing 0s into the MSBs of the native port, so that's probably bad, even if a subsequent axi write would *try* to increment the address by *half* a word size and populate those bits in a subsequent write... I'm going to leave this PR open for now, because I think it's still the right thing to do, but @enjoy-digital please sanity-check me before applying :)

gsomlo:
After some more thinking, I think the data-width assertion should go in. Also, I've added a commit introducing a base_address offset to the LiteDRAMNative2Wishbone converter, similar to the one recently added to LiteDRAMNative2AXI one.

enjoy-digital:
Thanks merged. It's indeed better to be strict on data_width between AXI and Native ports to avoid unspecified behavior.

PR #95, opened 2019-11-12
init: improve legibility of generated/sdram_phy.h (cosmetic)

gsomlo:
Use DFII_NPHASES constant (defined earlier in the file) to size the
two 'sdram_dfii_pix_[wrdata|rddata]_addr' arrays, instead of a bare
number. Additionally, ensure the first array element is indented by
a tab, same as the ones following it.


enjoy-digital:
Thanks, merged.

PR #104, opened 2020-01-24
phy/model: add support for initializing memory from file

### LINKED ISSUE #102: DFI Model: Add initialization capability
Currently, it's not possible to initialize the memory content of the DFI model at startup. Being able to initialize it at startup (similar to what we can already do for inferred memories or block rams), would allow simulating SoCs with firmware/OS binaries stored in the SDRAM and directly jump to it at startup from the bios, avoiding serialboot or tftp that are slow in simulation.

An `init` parameter should be added to `SDRAMPHYModel`, taking a list with the initialization datas (simular to Migen's `Memory`). Since the SDRAM is organized in banks, the initialization data then needs to be spread among banks to initialize each bank memory: https://github.com/enjoy-digital/litedram/blob/master/litedram/phy/model.py#L49

piotr-binkowski:
This PR adds support for initializing contents of `SDRAMPHYModel` which can greatly help with simulations.
This still needs some adjustments because LiteDRAM uses `ROW_BANK_COL` order and I initially went with simpler `BANK_ROW_COL` order. At this point data is visible in simulation but it is all over the place.

Closes #102

### DIFF
--- a/litedram/phy/model.py
+++ b/litedram/phy/model.py
@@ -19,7 +19,7 @@ from operator import or_
 # Bank Model ---------------------------------------------------------------------------------------
 
 class BankModel(Module):
-    def __init__(self, data_width, nrows, ncols, burst_length, we_granularity):
+    def __init__(self, data_width, nrows, ncols, burst_length, we_granularity, init):
         self.activate     = Signal()
         self.activate_row = Signal(max=nrows)
         self.precharge    = Signal()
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, before merging we should probably add an `address_mapping` parameter that defaults to `"ROW_BANK_COL"` and make sure this mapping is supported  since that's the mapping we will use for now.

piotr-binkowski:
I've made some improvements to this and it seems that now data in memory matches output from hexdump on a host PC, for some reason I needed to reverse the order in each 4 32-bit word group. Additionally it appears that at this point this won't work on memories with data bus wider than 32-bit because Migen seems to generate 32-bit wide init for 64-bit memories which then fails with `$readmem file address beyond bounds of array`

enjoy-digital:
Thanks for the changes and explanation, i'll do more tests with it and will try to remove the limitation.

PR #107, opened 2020-01-27
Count ACTIVATE commands in Bandwidth

jedrzejboczar:
This adds a register that counts activates so that we can measure bus efficiency.

As far as I understand it was not possible to refresh counters until period elapses, so software would have to constantly write to `update` and check if counters have been reset. I've changed the logic so that writing to `update` resets the counters for the sake of benchmarking. This allows to reset counters before starting benchmark, but this breaks bandwidth measurements. 
Should I create separate CSR that will be updated in this manner and leave the previous logic intact?

### DIFF
--- a/litedram/core/bandwidth.py
+++ b/litedram/core/bandwidth.py
@@ -51,10 +51,10 @@ class Bandwidth(Module, AutoCSR):
 
         self.sync += [
             Cat(counter, period).eq(counter + 1),
-            If(period,
-                nreads_r.eq(nreads),
-                nwrites_r.eq(nwrites),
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks @jedrzejboczar, sorry for not giving directions before, but for theses tests we are going to focus on simulation and will do the counting/observation directly in the simulation doing the benchmarking. For now it's also important to create tests around the code and avoid touching it. (we will be able to refactor/improve it when we'll have our unit-tests and benchmarks). I'm closing for now by maybe we will use it later or at least some part if we have a need to measure efficiency directly on the hardare.

PR #111, opened 2020-01-28
phy/model: simulate write latency

piotr-binkowski:
It appears that DDR2/DDR3 issues in enjoy-digital/litex#355 were caused by write data and write control signals being out of sync.
Screenshot_20200128_124005
After delaying the control signals connected to `BankModel` by `write_latency` cycles DDR2 and DDR3 memories work correctly in simulation


enjoy-digital:
I also did some tests on my side with SDR/DDR/LPDDR/DDR2 and DDR3 modules and memtest is always passing.

PR #115, opened 2020-01-28
phy/model: add burst_length value for ddr4 memories

piotr-binkowski:
It appears that only thing that prevents us from using this model with DDR4 memories is `burst_length` value, initially I've set it to 2 as this value is used by all DDR memory types.


enjoy-digital:
Thanks, i just created the issue with more information here: https://github.com/enjoy-digital/litex/issues/362 :) It should indeed work without other modification to the model.

PR #117, opened 2020-01-30
phy/model: init fixes

piotr-binkowski:
There are some limitations with current DRAM memory initialization, this PR aims to fix those issues
Currently known issues/limitations:

- [x] DRAM init works only on memories with 32-bit data bus
- [x] Each group of 4 32-bit words needs its order reversed
- [x] ~~Each group of n 128-bit words needs its order reversed (when using 256-bit data bus or wider, number of words must be tested)~~ ~~Not an issue since 128-bit bus support is enough~~ 64-bit wide DDR/DDR2 and 32-bit DDR3/DDR4 need this to work correctly
- [x] When using non-SDR modules data ends up on wrong offsets

### DIFF
--- a/litedram/phy/model.py
+++ b/litedram/phy/model.py
@@ -20,7 +20,7 @@ from operator import or_
 # Bank Model ---------------------------------------------------------------------------------------
 
 class BankModel(Module):
-    def __init__(self, data_width, nrows, ncols, burst_length, we_granularity, init):
+    def __init__(self, data_width, nrows, ncols, burst_length, nphases, we_granularity, init):
         self.activate     = Signal()
         self.activate_row = Signal(max=nrows)
         self.precharge    = Signal()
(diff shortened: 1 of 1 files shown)


piotr-binkowski:
I've managed to fix the issue with DDR data offsets - it was caused by simulation accessing only one bank model memory cell every (nphases*burst_length) ie. 8 in case of DDR3

It also appears that I will need to take care of the 256-bit issue because it seems to be triggered not by `-sdram-data-width` parameter but by bank model memory width

piotr-binkowski:
This fixes the issue with `init` working only for 32-bit wide SDR memory, it will now be possible to use `init` with other types of memory and different data widths

enjoy-digital:
Great, do you already have an idea for the:
`# Reverse order of 32-bit words in 128-bit groups` and
`# Reverse order of each 128-bit group`
If you have similar results for DDR2/DDR3/DDR4 with `--sdram-init` than with `--ram-init` with LiteX sim, we could merge this and i'll try to understand why the reversings are needed.

piotr-binkowski:
Tests with `--sdram-init` and `--ram-init` show that contents of the tested areas are the same and also match with the `init` file.
When it comes to the `reverse order` issue I just started looking at that and I wanted to start with checking how LiteDRAM controller generates addresses for the PHY and then how it handles the data on DFI bus.

enjoy-digital:
OK good, i'll just merge then and let you continue your understanding of the DFI bus.

Issue #119, opened 2020-01-30
Benchmark: add YAML configuration file


enjoy-digital:
The script executing the benchmark could take a YAML configuration file, similarly to LiteDRAM generator (https://github.com/enjoy-digital/litedram/blob/master/litedram/gen.py#L548).

The format could be:
````
{
    "test_0" : {
        "sdram_module":     "MT48LC16M16",
        "sdram_data_width": 32,
        "bist_length":      4096,
        "bist_random":      True
    },
    "test_1" : {
        "sdram_module":     "MT48LC16M16",
        "sdram_data_width": 32,
        "bist_length":      512,
        "bist_random":      False
    },
    "test_2" : {
        "sdram_module":     "MT46V32M16",
        "sdram_data_width": 32,
        "bist_length":      512,
        "bist_random":      False
    },
    [...]
}
````

PR #122, opened 2020-01-30
Load benchmark configuration from YAML

jedrzejboczar:
Closes https://github.com/enjoy-digital/litedram/issues/119.

The script now loads benchmarks configuration from a YAML file and allows to optionally filter tests based on their name using regex. 

I've put an example YAML in `test/`. Is it ok? Or maybe it should go to `examples/benchmarks`?

I also moved to relative imports in https://github.com/enjoy-digital/litedram/commit/096de78c63c5c5543453b014f830202774f098eb because `python setup.py test` was trying to load the file and failed due to import problem. This means that now the script has to be run with `python -m test.run_benchmarks <args>`. As another option we could change `python setup.py test` in Travis configuration to `python -m unittest`, which seems to avoid touching files that do not start with `test_`.

### DIFF
--- a/test/benchmarks.yml
+++ b/test/benchmarks.yml
@@ -1,0 +1,50 @@
+{
+    "test_0": {
+      "sdram_module":     'MT48LC16M16',
+      "sdram_data_width": 32,
+      "bist_length":      4096,
+      "bist_random":      True,
+    },
+    "test_1": {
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks! The location of benchmark.yml is fine, we will change it later if needeed when we'll have more feedback.

Issue #120, opened 2020-01-30
Benchmark: add graphs/plots of results


enjoy-digital:
With a preference for Matplotlib which is the most popular.

Issue #110, opened 2020-01-28
Benchmark: add benchmark script to Travis-CI


enjoy-digital:
Requires https://github.com/enjoy-digital/litedram/issues/109

The test should fail if any errors is detected during the test of the various configurations.

The results of the benchmark scripts should be stored somewhere to allow finding potential regressions in efficiency/performance on LiteDRAM.

jedrzejboczar:
With https://github.com/enjoy-digital/litedram/pull/123 merged, we could store the following artifacts:
* benchmark text summary (simple output redirection)
* plots
*  use `--output-cache` to store pickled pairs (benchmark_configuration, benchmark_output)

The pickled data can later be loaded by `run_benchmarks.py` without actually running the benchmarks, and so we can use it to generate output in different format if needed or to analyze LiteDRAM performance over time. 

If using `pickle`'s binary format is not ok, we can e.g. just store configurations and outputs in separate files, but pickling seems to be the easy solution.

mithro:
JSON seems like a better format than pickle?

enjoy-digital:
+1 for JSON.

PR #123, opened 2020-01-31
Add plotting of benchmark results

jedrzejboczar:
This adds option to plot the results of benchmarks using Matplotlib (closes https://github.com/enjoy-digital/litedram/issues/120).
The text summary has also been cleaned up a bit.

Plotting is enabled with `--plot` switch. 
By default plots are generated using non-GUI backed so `matplotlib` should be the only dependency when not using any GUI.

I've added an option to store intermediate results (output of individual benchmarks) in a file, so that it is possible to re-run the benchmarks fast with different summary (and possibly filtering results). It is stored using `pickle`.

So additional Python dependencies are `matplotlib` and `pickle`, but the script can run without them if not used. Should we list them in `setup.py` as [optional dependencies](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies)? Or maybe import them at the beginning and make them required?

A one-liner to quickly test available options:
```
python -m test.run_benchmarks --yaml test/benchmarks.yml --output-cache cache.pickle --regex 'test_[0-6]' --plot --plot-theme seaborn
```
Note that this will fail late if `matplotlib` or `pickle` is not installed, because they are imported only when needed. This might be inconvenient when running long benchmark only to find out that it fails at the end. Should I test the imports at the beginning and e.g. print warings if packages are not available?

### DIFF
--- a/test/run_benchmarks.py
+++ b/test/run_benchmarks.py
@@ -40,7 +40,7 @@ def human_readable(value):
 class BenchmarkConfiguration(Settings):
     def __init__(self, sdram_module, sdram_data_width, bist_length, bist_random):
         self.set_attributes(locals())
-        self._settings = {k: v for k, v in locals().items() if v != self}
+        self._settings = {k: v for k, v in locals().items() if k != 'self'}
 
     def as_args(self):
         args = []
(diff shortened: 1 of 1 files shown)


jedrzejboczar:
I've made the script exit with 1 if it finds any checker error (needed for https://github.com/enjoy-digital/litedram/issues/110)

enjoy-digital:
Thanks @jedrzejboczar, seems good! I'll test it and review it very soon.

enjoy-digital:
Merged and tested with small changes: https://github.com/enjoy-digital/litedram/commit/7e95ecc9a95b8a7aee855025e5d5704d43a2efbf and https://github.com/enjoy-digital/litedram/commit/736723cc98ae5605eb367bda73bf277375fd054b, the plots will be very useful!

PR #124, opened 2020-02-03
phy/model: cleanup the memory init code

### LINKED ISSUE #102: DFI Model: Add initialization capability

piotr-binkowski:
This PR removes code responsible for word order reversing as it is no longer needed since adding `l2_reverse` and additionally does a general cleanup by making a separate method from bank init code to only run it when needed

Closes #102

### DIFF
--- a/litedram/phy/model.py
+++ b/litedram/phy/model.py
@@ -113,6 +113,49 @@ class DFIPhaseModel(Module):
 # SDRAM PHY Model ----------------------------------------------------------------------------------
 
 class SDRAMPHYModel(Module):
+    def __prepare_bank_init_data(self, init, nbanks, nrows, ncols, data_width, address_mapping):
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, merged.

PR #125, opened 2020-02-03
phy/model: fix case when not converting init data width

piotr-binkowski:
This fixes an issue where init data was being overwritten when using configuration using 32-bit wide `BankModel` memories, ie. default SDRAM configuration.
This should help with latest issue from litex-hub/linux-on-litex-vexriscv#84


enjoy-digital:
Thanks, merged.

Issue #112, opened 2020-01-28
Benchmark: add Write/Read latency measurement


enjoy-digital:
Current benchmark does not measures Write/Read latency but only bandwidth/efficiency. To measure Write/Read latencies, `_LiteDRAMBISTGenerator` and `_LiteDRAMBISTChecker` ca, still be used, doing a single access. The latency will then be the reported `ticks`.

We must ensure `benchmark.py ` allows doing these tests from the command line and that the latency results are reported/plotted by the benchmark script (https://github.com/enjoy-digital/litedram/issues/109).

PR #127, opened 2020-02-03
Add Write/Read latency measurement

jedrzejboczar:
This adds option to measure read/write latency. Latency measurements are triggered by `--bist-length 1` and displayed/plotted separately.

Closes https://github.com/enjoy-digital/litedram/issues/112.

This also includes modification to use JSON instead of pickle as suggested in https://github.com/enjoy-digital/litedram/issues/110.


enjoy-digital:
Great! Merged, thanks.

PR #128, opened 2020-02-03
Add simulations to Travis CI

mglb:
This PR adds a new stage to Travis CI, which runs litex_sim and linux-on-litex-vexriscv simulations for each sdram module.

### DIFF
--- a/.sim-test.py
+++ b/.sim-test.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+import os
+import sys
+import pexpect
+import time
+from argparse import ArgumentParser
+
+
+parser = ArgumentParser()
+parser.add_argument("--sdram-module", type=str)
+args = parser.parse_args()
+
+
+tests = [
+    {
+        'id':      'litex_sim',
+        'command': f'litex_sim --with-sdram --sdram-module {args.sdram_module}',
+        'cwd':     os.getcwd(),
+        'checkpoints': [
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @mglb for the PR. just as a general idea, we have to avoid to much circular dependencies. LiteDRAM requires LiteX, Linux-on-LiteX-Vexriscv is an application project and requires LiteX / LiteDRAM (and others cores) be we should avoid LiteDRAM to require one of the application project. So we could run in Travis CI the tests based on LiteX sim, but the tests related to Linux-on-LiteX-Vexriscv should go probably go to https://github.com/litex-hub/linux-on-litex-vexriscv. If we want to be sure that changes to LiteDRAM are not breaking Linux-on-LiteX-Vexriscv, we should find a way to trigger Travis-CI of Linux-on-LiteX-Vexriscv when LiteDRAM is updated.

mglb:
I've removed linux simulations from this PR.

Each simulation test scans simulation output for memtest results. It should work as long as "BIOS built on" and "Memtest OK" strings are printed by the BIOS. Timeouts are at least 7x longer than it is usually needed.

enjoy-digital:
Thanks for the update, merged.

enjoy-digital:
@mglb: FYI, i removed the use of Conda with https://github.com/enjoy-digital/litedram/commit/8316ed3a47c8e402039e3804d8aa916cbe2f0e33. The setup is simple enough here to avoid using it (and i have a preference not using it when possible since it adds dependencies and potential problems).

PR #129, opened 2020-02-04
phy/model: add support for 8/16-bit wide SDR memory init

piotr-binkowski:
This PR adds support for initializing 8/16-bit SDR modules


enjoy-digital:
Thanks, looks good.

Issue #114, opened 2020-01-28
Benchmark: test simultaneous Write(s)/Read(s).


enjoy-digital:
In the current benchmark, Write and Read are done sequentially. We should add a test where Write and Read are done simultaneously (and be sure for this test that write datas are not overwriting expected read datas to avoid the checker to report false errors).

We should also make the number of generator and checker configurable to stress the core with multiple generators and checkers operating simultaneously.

Issue #113, opened 2020-01-28
Benchmark: Allow testing complex/pre-definied access patterns


enjoy-digital:
WIth current `_LiteDRAMBISTGenerator` and `_LiteDRAMBISTChecker` we can already test linear addressing or full random addressing (based on a LFSR). This will allow producing the first bandwidth/efficiency results for the different configurations, but we will probably need to test more specific access patterns when LiteDRAM is used for specific applications where access pattern is known in advance and we want to be sure to maximize performance. We should create a new gateware modules that are executing the access passed in the init parameters and with following prototypes:
````
address_data_pattern = [
    # address    data
    (0x00000001, 0x12345678),
    (0x00000010, 0xdeadbeef),
]

@ResetInserter()
class _LiteDRAMPatternGenerator(Module):
    def __init__(self, dram_port, init=[]):
        self.start  = Signal()
        self.done   = Signal()
        self.ticks  = Signal(32)

        # # #
[...]

@ResetInserter()
class _LiteDRAMPatternChecker(Module):
    def __init__(self, dram_port, init=[]):
        self.start  = Signal()
        self.done   = Signal()
        self.ticks  = Signal(32)

        # # #

[...]

pattern_generator = _LiteDRAMPatternGenerator(self.sdram.crossbar.get_port(), init=address_data_pattern)
pattern_checker   = _LiteDRAMCheckerGenerator(self.sdram.crossbar.get_port(), init=address_data_pattern)
````

And integrate them in the benchmark. We should be able to pass the address/data pattern from the command line: `--pattern-init=...`

Theses modules could be located in `frontend/bist.py`.

PR #131, opened 2020-02-04
Allow testing custom access patterns

jedrzejboczar:
PR for issue https://github.com/enjoy-digital/litedram/issues/113.
It adds possibility to run benchmarks using predefined access patterns loaded from simple CSV files. 

It is WIP as currently it is not possible to use custom patterns for `run_benchmarks.py`. I have to modify the way summary is generated to account for cases of custom access pattern, simple latency benchmarks and simultaneous write+read (https://github.com/enjoy-digital/litedram/issues/114). But `run_benchmarks.py` will still work correctly if custom access patterns are not used.

As for the `benchmark.py`, currently there is a limitation that the access pattern cannot write to the same address twice, because this will result in data being overwritten before checker reads them. If needed, this will be easier to implement after https://github.com/enjoy-digital/litedram/issues/114.

I also added two simple helper scripts for generating configuration and access pattern, to this PR can e.g. be run using:
```
./benchmark.py --access-pattern <(./gen_access_pattern.py 0 1024 32)
```


enjoy-digital:
Thanks, this seems fine. In the future, we'll want to be able to use the pattern generator/checker on hardware, so instead of the storing the data/address with `Case` that would be better to use a `Memory`.

jedrzejboczar:
I've rewritten results processing to be more flexible. The benchmark results can have quite different parameters depending on configuration, so to I used [pandas](https://pandas.pydata.org/pandas-docs/version/1.0.0/whatsnew/v1.0.0.html) to keep results processing simpler. Results are now displayed in groups depending on access type to make them comparable. 

Configuration file format has changed a bit to allow for custom access patterns from file. Access pattern cannot be stored directly in the configuration, but we need some name to display it anyway, so we just store access patterns in files and display the name of the file.

jedrzejboczar:
Could we marge this one now? I will be creating a PR related to speeding up the benchmark on top of this one and later I will implement access patterns using `Memory` in a separate PR.

enjoy-digital:
OK it's merged, i just created: https://github.com/enjoy-digital/litedram/issues/134.

PR #135, opened 2020-02-07
Improve speed of benchmark runner

jedrzejboczar:
These changes should significantly speed up the process of running benchmarks and generating summary:
1. We can run benchmarks as parallel jobs using `multiprocessing`, by default `os.cpu_count()` is used, but it can be controller with `--njobs`.
2. By default, we continue running even if some benchmarks fail, and a summary of failures will be displayed at the end. `--fail-fast` can be specified if we want to exit on any error.
3. Removed the code that instantiated `LiteDRAMBenchmarkSoC`, as it was taking a lot of time to do it for each configuration.
4.  I've fixed problem with plots being unreadable when there are too many entries; now we just increase figure size as needed.

About 3:
`LiteDRAMBenchmarkSoC` was only used to retrieve clock frequency and interface data width. Now data width is just calculated from SDRAM module class and the clock frequency is hardcoded as it is done [here](https://github.com/enjoy-digital/litex/blob/1d70ef6958bfc11cafc80d29e0ad5550d9d0fc89/litex/tools/litex_sim.py#L184) in `litex_sim.py`. 
Maybe we could avoid hard coding it, by exposing this frequency as a global variable in `litex_sim.py` and importing it here? This would at least fail fast if the variable was removed in `litex_sim.py`, but maybe there exists a better solution.

### DIFF
--- a/test/run_benchmarks.py
+++ b/test/run_benchmarks.py
@@ -28,13 +28,6 @@ from . import benchmark
 from .benchmark import LiteDRAMBenchmarkSoC, load_access_pattern
 
 
-def center(text, width, fillc=' '):
-    added = width - len(text)
-    left = added // 2
-    right = added - left
-    return fillc * left + text + fillc * right
-
-
 # Benchmark configuration --------------------------------------------------------------------------
 
 class Settings(_Settings):
@@ -174,7 +167,7 @@ class BenchmarkResult:
     def find(pattern, output):
         result = pattern.search(output)
         assert result is not None, \
-            'Could not find pattern "%s" in output:\n%s' % (pattern, benchmark_output)
+            'Could not find pattern "%s" in output' % (pattern)
         return int(result.group('value'))
 
     def __init__(self, output):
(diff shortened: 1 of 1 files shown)


test/run_benchmarks.py
```diff
@@ -488,6 +523,7 @@ def main(argv=None):
     if _summary:
         summary = ResultsSummary(run_data)
         summary.text_summary()
+        summary.failuers_summary()
```
enjoy-digital: small typo: failuers ->failures

enjoy-digital:
Thanks, this is fine to have some hardcoded parameters for now, can you just list the limitations in the file header so that we don't forget? something like:
````
Limitations/TODO:
- add configurable data_width?
- add configurable sys_clk_freq?
````

There is a small typo, failuers --> failures. Can you fix that? we could then merge.

jedrzejboczar:
Sure, I've fixed these now.

enjoy-digital:
Thanks, merged!

Issue #134, opened 2020-02-06
LiteDRAMPatternGenerator/Checker: use Memory instead of Case


enjoy-digital:
As discussed in https://github.com/enjoy-digital/litedram/pull/131#issuecomment-581986772.

PR #136, opened 2020-02-07
Benchmark: use Memory instead of Case for custom access pattern

jedrzejboczar:
Closes https://github.com/enjoy-digital/litedram/issues/134


jedrzejboczar:
And probably we can also close https://github.com/enjoy-digital/litedram/issues/113

enjoy-digital:
Thanks, merged.

Issue #133, opened 2020-02-05
Benchmark: BIST not generating random address


jedrzejboczar:
`random` register in `_LiteDRAMBISTGenerator` and `_LiteDRAMBISTChecker` is used for random data generation, but address is always sequential:
https://github.com/enjoy-digital/litedram/blob/master/litedram/frontend/bist.py#L145

Random data has no impact on LiteDRAM performance, shouldn't we generate random address instead? Or add separate registers for random data and random address?

enjoy-digital:
`random` was used to test integrity of the data, this has indeed no impact on the efficiency/performance. We could add two separate CSR: `random_data` and `random_addr`. But we should probably add another CSR to limit the range of the random address, so maybe add a `end` CSR (similar to `base)` and random addressing would be done between `base` and `end.`

jedrzejboczar:
Is there any good way of limiting the generated value to a range when `end` is a CSR? Because I cannot think of a nice solution.

It would be the easiest if `base` and `end` were just parameters passed to `__init__`, even better if we only allowed `end - base` to be a power of 2. Or just pass `range_bits`. We could then just create an LSFR with desired output width.

When `end` is a CSR with arbitrary address, we may check if `generated > end` and if not, then choose some other value. This could be just `base`/`end`, or we could switch to a simple counter for as long as we get random values out of range. But this will give very bad results for small ranges, because we would then almost always generate values out of range, so almost all values would be from the fallback strategy.

enjoy-digital:
In fact we need `base` and `end` to be dynamic (so CSR) for the hardware tests were we will want to re-configure `_LiteDRAMBISTGenerator` and `_LiteDRAMBISTGenerator:`

We can do the assumption that the configured range will always be a power of 2, and then you can probably do something similar to:

```
addr_gen = Generator(31, n_state=31, taps=[27, 30]) # PRBS31
addr_mask = Signal(32)
self.comb += addr_mask.eq(self.end.storage - self.base.storage)
self.comb += dma.sink.address.eq(self.base.storage + addr_gen.o & addr_mask)
```

jedrzejboczar:
Thanks, that's nice and simple, I'll implement it this way.

PR #137, opened 2020-02-11
Benchmark: add simultaneous Write/Read and random address generation

jedrzejboczar:
Closes https://github.com/enjoy-digital/litedram/issues/133.
Partially implements https://github.com/enjoy-digital/litedram/issues/114 (still single generator and checker).

I've added random address generation as suggested in https://github.com/enjoy-digital/litedram/issues/133, but limiting values makes them not unique, so to actually use such generator, I had to also implement simultaneous write/read.

To allow for alternating write/read patterns, I added `run` and `ready` signals to generators and checkers. When `run` is low, generator will stop writing/reading and go to `WAIT` state until `run` is high again. To have the previous behaviour, we can just set `run.eq(1)`, and by cross-connecting the signals between generator and checker, we can make them perform alternating write/read. 

Currently it is still not possible to run non-alternating write/read with random addressing, as the data will get overwritten before checker reads it, so there is an assertion in `benchmark.py`.

I've also updated benchmark configuration, runner and summary generation to include these new options.

### DIFF
--- a/litedram/frontend/bist.py
+++ b/litedram/frontend/bist.py
@@ -133,6 +133,8 @@ class _LiteDRAMBISTGenerator(Module):
         ashift, awidth = get_ashift_awidth(dram_port)
         self.start  = Signal()
         self.done   = Signal()
+        self.run    = Signal()
+        self.ready  = Signal()
         self.base   = Signal(awidth)
         self.length = Signal(awidth)
         self.random = Signal()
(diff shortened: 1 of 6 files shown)


enjoy-digital:
Thanks, LGTM! I'll try to see in the future if we can have unique values generated with the generator and eventually avoid the `run`/`ready` signals, but for now this is a good workaround/compromise.

PR #140, opened 2020-02-12
Benchmark: use variable number of BIST generators/checkers

jedrzejboczar:
Closes https://github.com/enjoy-digital/litedram/issues/114.

This adds an option to specify `--num-generators` and `--num-checkers` for `benchmark.py` to have more than one generator/checker. 

Max value of ticks is used as the final benchmark result (for generator ticks, checker ticks and checker errors). In the final bandwidth/efficiency calculations the number of data is multiplied according to the number of generators/checkers.

I've also updated `gen_config.py`. It is useful for testing purpose, but product of all options quickly results in over 1000 benchmarks, so for real benchmarks we should most likely specify configurations by hand anyway.

### DIFF
--- a/litedram/frontend/bist.py
+++ b/litedram/frontend/bist.py
@@ -324,17 +324,17 @@ class LiteDRAMBISTGenerator(Module, AutoCSR):
     """
     def __init__(self, dram_port):
         ashift, awidth = get_ashift_awidth(dram_port)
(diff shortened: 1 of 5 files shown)


enjoy-digital:
Great, thanks. That's merged. We will indeed probably have to limit the configurations/tests we want to run.

PR #141, opened 2020-02-13
Benchmark: reduce disk usage when running benchmarks in parallel

jedrzejboczar:
This is to solve the problem of unbound disk usage when running benchmarks in parallel. Previously each benchmark case was being run in a separate directory, but this does not scale well when running a lot of benchmarks (~20MB/benchmark). Now we use one directory per process, so that build directories are reused.


enjoy-digital:
Thanks, merged.

PR #142, opened 2020-02-14
Update tREFI verifier

piotr-binkowski:
This PR updates the `tREFI` verifier to be less strict and check only if average `tREFI` constraint was satisfied. Previous test that checked if there was at most `tREFI` delay between `REF` commands is only enabled in verbose mode.
Additionally this adds checks for refresh postponing that is supported from DDR onwards.

It is also worth noting that this seems to fix the `tREFI` issues mentioned in litex-hub/linux-on-litex-vexriscv#103 which could mean that the current model is too strict


enjoy-digital:
Thanks, we can merge it. I'll review it more carefully when doing a full review of the Timing Checker.

PR #143, opened 2020-02-17
phy/model: fix memory addressing issues in some configurations

piotr-binkowski:
This fixes litex-hub/linux-on-litex-vexriscv#104

### DIFF
--- a/litedram/phy/model.py
+++ b/litedram/phy/model.py
@@ -62,13 +62,13 @@ class BankModel(Module):
         rdaddr         = Signal(max=bank_mem_len)
 
         self.comb += [
-            wraddr.eq(row*ncols | self.write_col),
-            rdaddr.eq(row*ncols | self.read_col),
+            wraddr.eq((row*ncols | self.write_col)[log2_int(burst_length*nphases):]),
+            rdaddr.eq((row*ncols | self.read_col)[log2_int(burst_length*nphases):]),
         ]
 
         self.comb += [
             If(active,
-                write_port.adr.eq(wraddr[log2_int(burst_length*nphases):]),
+                write_port.adr.eq(wraddr),
                 write_port.dat_w.eq(self.write_data),
                 If(we_granularity,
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks.

enjoy-digital:
@piotr-binkowski: Travis CI is now failing on Linux-on-LiteX-VexRiscv for some configurations: https://github.com/litex-hub/linux-on-litex-vexriscv/runs/453115052 I haven't look closely yet, but do you think it could be related to this change? (that's the last change we did that i think of and that could explain the regression).

piotr-binkowski:
I think that this might be related, from what I see this only affects chips with `nrows  = 65536` and I've modified part of the code that handles addressing. I will look into this

enjoy-digital:
OK thanks.

piotr-binkowski:
I think that I found the issue. It appears that there is another thing that those memories have in common - each one of them is larger than 1GB when used with the default data width(32-bit).
Simulations showed that when CPU was accessing `0x40000000`(DRAM offset 0) LiteDRAM sends read requests for row `0x8000` instead of the expected `0x0000` - somehow already accessing the upper half of the memory while Litex and BIOS state that this memory is only 1 GB in size.
@enjoy-digital could you please take a look at this issue because I think that while my change surfaced the issue in CI, the issue itself might be located in another part of the LiteDRAM core and I'm not sure where to look for it.

enjoy-digital:
Thanks for looking at the issue @piotr-binkowski, i'll continue the investigation.

enjoy-digital:
The issue was that the `base_address` or `main_ram` was not substracted. This was causing issue on large memories since the MSB were passed to the DRAM controller (which was not the case on regular memories since the MSB was truncated). This is fixed with https://github.com/enjoy-digital/litex/commit/f3829cf0819c2e5d0ebc6cffb9569f9f46191549 and this is the first issue found with the new features/tests :)

PR #146, opened 2020-02-17
Benchmarks: Generate HTML summary and deploy it from Travis

jedrzejboczar:
This PR adds benchmarks summary in HTML format and deployment to GitHub Pages.
Closes https://github.com/enjoy-digital/litedram/issues/110.

**Travis:**

Benchmarks are being run in Travis and deployed using GitHub Pages. The results are stored on the `gh-pages` branch. We would then need to [enable GitHub Pages in repo settings](https://help.github.com/en/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site) for `enjoy-digital/litedram` and configure it to use `gh-pages` branch. The results should then appear at [enjoy-digital.github.io/litedram/](https://enjoy-digital.github.io/litedram/).

[Here](https://docs.travis-ci.com/user/deployment/pages/#setting-the-github-token) is some more information about Travis/gh-pages. 
Options used in `.travis.yml`:
* `deploy.keep_history = true` - instead of force pushing to the repository we have commits with previous versions of the summaries, this way we'll be able to compare results across history
* `deploy.on.branch = master` - deployment is skipped on branches other than master; or maybe we could entirely skip benchmarks on other branches?
* `deploy.token` - secure variable `GITHUB_TOKEN` is being used, you would need to [generate it](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) (need only `public_repo`) and set it as [repository variable](https://docs.travis-ci.com/user/environment-variables#defining-variables-in-repository-settings)

**HTML summary:**

The generated summary consists only of tables with results (example summary: [summary.zip](https://github.com/enjoy-digital/litedram/files/4214090/summary.zip)). As for now it doesn't include plots. Tables are probably more practical, as we have options to sort by whichever column we want and we have much more information available. Plots show results by test name, which is often meaningless without looking at the YAML configuration, but it is not easy to include the configuration in the plots. Maybe by switching to something like [Plotly](https://plot.ly/javascript/) and adding dynamic legend on hover.

During deployment we store the HTML page and results cache JSON file. We could maybe store SVG plots too, although they can be generated using the JSON cache.

The generated HTML page includes the time of generation and `litedram` commit hash, but maybe we should also store this version information in a separate file as I could not find a way to provide custom commit message.

Additionally `b7ed91d` changes logging level in `benchmark.py` to `WARNING` as the logs are displayed on standard error which is not captured by `run_benchmarks.py`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,15 @@ python: "3.6"
 before_install:
   - sudo apt-get update
   - sudo apt-get -y install verilator libevent-dev libjson-c-dev
-  - pip install pexpect
+  - pip install pexpect numpy matplotlib pandas jinja2
 
 install:
   # Get Migen / LiteX / Cores
   - wget https://raw.githubusercontent.com/enjoy-digital/litex/master/litex_setup.py
   - python3 litex_setup.py init install
+  # litex_setup.py fails for litedram, which has already been cloned by Travis
+  # now install the local litedram so that it is available for benchmarks script
+  - python3 -m pip install .
 
 before_script:
   # Get RISC-V toolchain
@@ -58,3 +61,19 @@ jobs:
     - env: SDRAM_MODULE=MT40A1G8
     - env: SDRAM_MODULE=MT40A512M16
 
+    - stage: Benchmarks
+      script:
+      - python3 -m test.run_benchmarks test/benchmarks.yml --results-cache cache.json --html
+      # move benchmark artifacts to gh-pages/ directory that will be pushed to gh-pages branch
+      - mkdir -p gh-pages
+      - mv html/summary.html gh-pages/index.html
+      - mv cache.json gh-pages/cache.json
+      - touch gh-pages/.nojekyll
+      deploy:
+        provider: pages
+        skip_cleanup: true
+        token: $GITHUB_TOKEN
+        keep_history: true
+        local_dir: gh-pages
+        on:
+          branch: master
--- a/test/benchmark.py
+++ b/test/benchmark.py
@@ -4,6 +4,7 @@
 # License: BSD
 
 import csv
+import logging
 import argparse
 from operator import and_
 from functools import reduce
@@ -219,7 +220,12 @@ def main():
     parser.add_argument("--num-generators",   default=1,              help="Number of BIST generators")
     parser.add_argument("--num-checkers",     default=1,              help="Number of BIST checkers")
     parser.add_argument("--access-pattern",                           help="Load access pattern (address, data) from CSV (ignores --bist-*)")
+    parser.add_argument("--log-level",        default="info",         help="Set logging verbosity",
+                        choices=['critical', 'error', 'warning', 'info', 'debug'])
     args = parser.parse_args()
+
+    root_logger = logging.getLogger()
+    root_logger.setLevel(getattr(logging, args.log_level.upper()))
 
     soc_kwargs     = soc_sdram_argdict(args)
     builder_kwargs = builder_argdict(args)
--- a/test/run_benchmarks.py
+++ b/test/run_benchmarks.py
@@ -492,7 +492,8 @@ def run_single_benchmark(func_args):
     # run as separate process, because else we cannot capture all output from verilator
     print('  {}: {}'.format(config.name, ' '.join(config.as_args())))
     try:
-        output = run_python(benchmark.__file__, config.as_args() + ['--output-dir', output_dir])
+        args = config.as_args() + ['--output-dir', output_dir, '--log-level', 'warning']
+        output = run_python(benchmark.__file__, args)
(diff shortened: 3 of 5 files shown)


enjoy-digital:
Great, thanks. Storing only the tables seems fine for now if we are still able to re-generate the plots locally. In the future, doing the plots dynamically from the tables would indeed probably be a better option than storing the generated plots.

enjoy-digital:
@jedrzejboczar: it seemed to have failed when deploying the results: https://travis-ci.org/enjoy-digital/litedram/jobs/651954280?utm_medium=notification&utm_source=github_status can you check if you understand and if i did something wrong when enabling Github Pages?

jedrzejboczar:
[gh-pages](https://github.com/enjoy-digital/litedram/tree/gh-pages) branch has been initialized and can be accessed at [enjoy-digital.github.io/litedram](http://enjoy-digital.github.io/litedram/). The error is most likely due to missing `GITHUB_TOKEN` variable: 
https://github.com/enjoy-digital/litedram/blob/5fb2b011d88e5fc1438de2b2e031431e89935d56/.travis.yml#L75

You would need to [generate an access token](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) (only `public_repo` access is needed) and [set it as repository variable](https://docs.travis-ci.com/user/environment-variables#defining-variables-in-repository-settings) with name `GITHUB_TOKEN`. Then I believe that everything should work fine.

enjoy-digital:
Thanks, that's now configured. It seems something is going wrong with the benchmark on Travis-CI: https://travis-ci.org/enjoy-digital/litedram/jobs/651954280?utm_medium=notification&utm_source=github_status do you have an idea of what is going on?

jedrzejboczar:
I think benchmarks are being run correctly and the problem is with their output. There seems to be some kind of buffering of `stdout`. The warnings from build process on `stderr` are printed correctly, but the messages `test_X: ...` that should appear before the `benchmark.py` is even started are not visible.

I will try to test it with those messages being also printed to `stderr` and will create a new PR if this helps. 

I would like to avoid printing some artificial "hearbeat" messages in `run_benchmarks.py`, as then the build would not be stopped if `benchmark.py` does not finish for some reason, but starts the BIOS prompt. But maybe it would be good idea to add a timeout on all `benchmark.py` runs anyway, so that we prevent such situations even outside of CI. Then the "heartbeat" could be directly in `run_benchmarks.py`.

jedrzejboczar:
In one of my successful runs the output looks like this, so there is some kind of buffering and `stdout` is only printed after benchmarks are finished:
```
Using 2 parallel jobs
/home/travis/build/antmicro/litedram/litex/litex/soc/software/compiler_rt/lib/builtins/comparesf2.c:85:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
 FNALIAS(__cmpsf2, __lesf2);
 ^~~~~~~
/home/travis/build/antmicro/litedram/litex/litex/soc/software/compiler_rt/lib/builtins/comparesf2.c:85:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
 FNALIAS(__cmpsf2, __lesf2);
 ^~~~~~~
/home/travis/build/antmicro/litedram/litex/litex/soc/software/compiler_rt/lib/builtins/comparedf2.c:85:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
 FNALIAS(__cmpdf2, __ledf2);
 ^~~~~~~
/home/travis/build/antmicro/litedram/litex/litex/soc/software/compiler_rt/lib/builtins/comparedf2.c:85:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
 FNALIAS(__cmpdf2, __ledf2);
 ^~~~~~~
  test_0: --sdram-module=MT41K128M16 --sdram-data-width=32 --num-generators=1 --num-checkers=1 --bist-alternating --bist-length=1 --bist-random
  test_0: ok
  test_3: --sdram-module=MT41K128M16 --sdram-data-width=32 --num-generators=1 --num-checkers=1 --bist-alternating --bist-length=1024
  test_3: ok
  test_1: --sdram-module=MT41K128M16 --sdram-data-width=32 --num-generators=1 --num-checkers=1 --bist-alternating --bist-length=1
  test_1: ok
  test_2: --sdram-module=MT41K128M16 --sdram-data-width=32 --num-generators=1 --num-checkers=1 --bist-alternating --bist-length=1024 --bist-random
  test_2: ok
===> Latency:
     name sdram_module  sdram_data_width  bist_alternating  num_generators  num_checkers write_latency read_latency
0  test_0  MT41K128M16                32              True               1             1         2 clk       36 clk
1  test_1  MT41K128M16                32              True               1             1         2 clk       36 clk
```

PR #148, opened 2020-02-19
Benchmark: Print heartbeat message during runs in Travis CI

jedrzejboczar:
This PR adds `--heartbeat` argument that is being used in CI to print heartbeat messages every 60 seconds. 

This solves the issue mentioned in https://github.com/enjoy-digital/litedram/pull/146#issuecomment-588089239. After some tests I've found that we need to print in a child process instead of the main process, as both `stdout` and `stderr` from main process seem to be buffered by Travis. This behavior can be reproduced using:
```
python -m test.run_benchmarks test/benchmarks.yml --njobs 2 --regex 'test_[0-3]$' >/tmp/output 2>&1
```
and then reading the file, e.g. `tail -f /tmp/output`.

Another option would be to print something on `stderr` in `benchmark.py` (e.g. `argv`), as with the current solution, because of Travis buffering, we don't get real-time feedback which tests is currently being run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ jobs:
 
     - stage: Benchmarks
       script:
-      - python3 -m test.run_benchmarks test/benchmarks.yml --results-cache cache.json --html
+      - python3 -m test.run_benchmarks test/benchmarks.yml --results-cache cache.json --html --heartbeat 60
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks merged.

PR #152, opened 2020-02-20
Benchmarks: add timeout parameter

jedrzejboczar:
Travis normally times out a build after no output for 10 minutes. We've added a "heartbeat" to prevent this from canceling benchmarks, but it would be nice to have a timeout for some situations (e.g. when `$finish` is somehow never executed). 

In this PR I've added a `--timeout` parameter for CI builds. It seems that a single benchmark case takes ~30s, so the timeout of 9 minutes that I've added is somehow overestimated, if you want we can change it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ jobs:
 
     - stage: Benchmarks
       script:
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks. For now the 9 minutes are fine, we could reduce it in the future if needed.

PR #153, opened 2020-02-20
test/benchmarks: add memtype to summary (#151)

### LINKED ISSUE #151: Benckmark: Add SDRAMModule's memtype to results
While navigating in http://enjoy-digital.github.io/litedram, it would be useful to have a memtype column, since results greatly depends of this. This would allow easy comparison between different SDRAM module of the same memtype. cc @jedrzejboczar.

jedrzejboczar:
Closes #151.


enjoy-digital:
Thanks.

PR #154, opened 2020-02-26
modules: add KVR21SE15S8/4 SO-DIMM

piotr-binkowski:
This PR adds definition for a module that I'm using in litex-hub/litex-boards#47


enjoy-digital:
Thanks, merged.

PR #157, opened 2020-03-05
Support for multiple SDRAM PHYs in single SoC

mglb:
Requires https://github.com/enjoy-digital/litex/pull/411. See the Litex PR for details.


enjoy-digital:
Closing, we'll continue the discussion in: https://github.com/enjoy-digital/litex/issues/346. In the future I would appreciate being able to discuss these features before too much time is spent in the implementation, since here it was probably too early for the project and could probably have been worked around for the Pano Logic G2.

Issue #156, opened 2020-03-05
Add unit-tests to litedram.frontend modules


enjoy-digital:
In order to be able to refactor and rework the frontend modules of LiteDRAM, we first need to improve unit-tests of these modules.

Some of the frontend modules have unit-tests with good coverage of the functionality (ex AXI) but others are really minimal or just checking the syntax. We should make sure all frontend modules are well tested:

**Adaptation**
- [x] litedram.frontend.adaptation.LiteDRAMNativePortCDC
- [x] litedram.frontend.adaptation.LiteDRAMNativePortDownConverter
- [x] litedram.frontend.adaptation.LiteDRAMNativeWritePortUpConverter

**AXI**
- [x] litedram.frontend.axi.LiteDRAMAXI2NativeW
- [x] litedram.frontend.axi.LiteDRAMAXI2NativeR
- [x] litedram.frontend.axi.LiteDRAMAXI2Native

**BIST**
- [x] litedram.frontend.bist.Generator
- [x] litedram.frontend.bist._LiteDRAMBISTGenerator
- [x] litedram.frontend.bist._LiteDRAMPatternGenerator
- [x] litedram.frontend.bist.LiteDRAMBISTGenerator
- [x] litedram.frontend.bist._LiteDRAMBISTChecker
- [x] litedram.frontend.bist._LiteDRAMPatternChecker
- [x]  litedram.frontend.bist.LiteDRAMBISTChecker

**DMA**
- [x] litedram.frontend.dma.LiteDRAMDMAReader
- [x] litedram.frontend.dma.LiteDRAMDMAWriter

**ECC**
- [x] litedram.frontend.ecc.LiteDRAMNativePortECCW
- [x] litedram.frontend.ecc.LiteDRAMNativePortECCR
- [x] litedram.frontend.ecc.LiteDRAMNativePortECC

**FIFO**
- [x] litedram.frontend.fifo._LiteDRAMFIFOCtrl
- [x] litedram.frontend.fifo._LiteDRAMFIFOWriter
- [x] litedram.frontend.fifo._LiteDRAMFIFOReader
- [x] litedram.frontend.fifo.LiteDRAMFIFO

**Wishbone**
- [x] litedram.frontend.wishbone.LiteDRAMWishbone2Native

enjoy-digital:
Note: Random valid/ready is not tested on LiteDRAMDMAReader/LiteDRAMDMAWriter.

enjoy-digital:
[...]

PR #161, opened 2020-03-11
test: add _LiteDRAMBISTGenerator tests

jedrzejboczar:
I'm working on https://github.com/enjoy-digital/litedram/issues/156. 

This PR only adds some some tests for `_LiteDRAMBISTGenerator` and I will continue to write tests for other classes, but I wanted to create the PR now, so that you can comment if it's ok or maybe you would prefer to start from other modules?

### DIFF
--- a/test/test_bist.py
+++ b/test/test_bist.py
@@ -29,9 +29,11 @@ class GenCheckDriver:
         yield self.module.reset.eq(0)
         yield
 
-    def run(self, base, length):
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, this is fine and we can merge it. You can start with modules you are more comfortable with or have a better understanding, there are no order for this, the important thing in the end is to have a good coverage of each module.

PR #162, opened 2020-03-12
Benchmarks: fix with_uart parameter

jedrzejboczar:
Due to https://github.com/enjoy-digital/litex/commit/b29f443fe51511df908a644a17bc2a8a1f1566cc removing the default parameter value, we need to set it explicitly.

To prevent something like this from being unnoticed (e.g. https://travis-ci.org/github/enjoy-digital/litedram/jobs/661098966?utm_medium=notification&utm_source=github_status succeeds) `run_benchmarks.py` will now fail in case all benchmarks failed.

### DIFF
--- a/test/benchmark.py
+++ b/test/benchmark.py
@@ -235,6 +235,8 @@ def main():
     sim_config.add_module("serial2console", "serial")
 
     # Configuration --------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks merged.

PR #163, opened 2020-03-13
test: add tests for _LiteDRAMPatternGenerator

jedrzejboczar:
While adding tests for `_LiteDRAMPatternGenerator`, I've noticed an inconsistency with `_LiteDRAMBISTGenerator`. The address list passed in `_LiteDRAMPatternGenerator`'s `init` is not interpreted as actual address, but as word number. This is different than `base` and `end` in `_LiteDRAMBISTGenerator`, where it is used as byte address (and auto aligned to word size). 

We should probably make it behave the same as `_LiteDRAMBISTGenerator` by changing https://github.com/enjoy-digital/litedram/blob/60b618eebaebabeea07f9d1082653d1888fb89f4/litedram/frontend/bist.py#L280 to `dma_sink_addr.eq(addr_port.dat_r[ashift:])`.

I would do that after adding tests for `_LiteDRAMPatternChecker`, because this change would also require modifying benchmarks code to generate properly aligned access patterns (to avoid duplicates which cause checker errors without `--bist-alternating`).

### DIFF
--- a/test/test_bist.py
+++ b/test/test_bist.py
@@ -29,12 +29,15 @@ class GenCheckDriver:
         yield self.module.reset.eq(0)
         yield
 
-    def run(self, base, length, end=None):
+    def configure(self, base, length, end=None):
+        # for non-pattern generators/checkers
         if end is None:
             end = base + 0x100000
         yield self.module.base.eq(base)
         yield self.module.end.eq(end)
         yield self.module.length.eq(length)
+
+    def run(self):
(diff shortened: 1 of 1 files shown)


jedrzejboczar:
As I've added some more tests now, I'm closing this PR and will create a bigger one for all bist/pattern generators/checkers. I'm also moving the addressing problem to an issue (https://github.com/enjoy-digital/litedram/issues/164).

Issue #164, opened 2020-03-13
Use consistent addressing in BIST/Pattern generators


jedrzejboczar:
In `_LiteDRAMBISTGenerator` the `base` address is being aligned to `dram_port.data_width` and `length` denotes the number of commands issued (https://github.com/enjoy-digital/litedram/blob/master/litedram/frontend/bist.py#L187).

In `_LiteDRAMPatternGenerator` the address passed in `init` is not being shifted. We should probably treat it the same as in `_LiteDRAMBISTGenerator` by changing https://github.com/enjoy-digital/litedram/blob/60b618eebaebabeea07f9d1082653d1888fb89f4/litedram/frontend/bist.py#L280 to `dma_sink_addr.eq(addr_port.dat_r[ashift:])`.

Changing this will also require updating benchmarks to generate proper access patterns depending on `dram_port` data width. I'll work on this after adding some more BIST unit tests.

PR #165, opened 2020-03-13
Test: add tests for BIST modules with different access patterns

jedrzejboczar:
This PR adds tests for `_LiteDRAMBISTGenerator`, `_LiteDRAMPatternGenerator`, `_LiteDRAMBISTChecker`, `_LiteDRAMPatternChecker`. Test data definitions are created in `setUp` and then used as reference in different test.

It tests current pattern addressing (not https://github.com/enjoy-digital/litedram/issues/164).

If you would prefer to wait for all the BIST related tests, than I can just commit to this PR.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -1,5 +1,6 @@
 # This file is Copyright (c) 2016-2019 Florent Kermarrec <florent@enjoy-digital.fr>
 # This file is Copyright (c) 2016 Tim 'mithro' Ansell <mithro@mithis.com>
+# This file is Copyright (c) 2020 Antmicro <www.antmicro.com>
 # License: BSD
 
 import random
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks. If you still have tests you want to do for the BIST and it's not blocking you if this PR is not merged for now, you can add then to this PR yes. I'll try to review it soon.

jedrzejboczar:
I've added tests with memory using clock domain asynchronous to `sys` and it seems that there is a problem with synchronization in BIST modules.

`BusSynchronizer`s take much more time for `base` and `error` signals to propagate, then for `start` and `done` signals, so e.g. a write to `base` followed by a write to `start` results in the first word being written by generator using old value of `base` before the new value propagates through the synchronization logic.

This of course can be fixed by adding delays, but I wonder if maybe we should add some kind of synchronization on `start` and `done` signals that ensures that all other values get propagate before.

**More information:**

The tests have the same [sequence](https://github.com/antmicro/litedram/blob/029fbed25e5113504add1c3ea2a1a213de5109a6/test/test_bist.py#L483) as before - write, read, write to another location (corrupt few values written previously), read first location (errors), read new locaton (ok). With this setup [test_bist_csr_cdc](https://github.com/antmicro/litedram/blob/029fbed25e5113504add1c3ea2a1a213de5109a6/test/test_bist.py#L562) fails due to:
- 2nd checker run reports 1 error; ideally it should be 4, but due to generator writing first word to wrong location there are actually 3 errors 
- 3rd checker run reports 1 error (at address 0x10, which is where generator written value because `base` was not yet updated to new value)

If you want to take a look, I attach a gtkwave file for that test (removed asserion for errors == 4 to have full simulatin): [sim-gtkwave.zip](https://github.com/enjoy-digital/litedram/files/4339151/sim-gtkwave.zip)

jedrzejboczar:
I've added the remaining BIST tests from https://github.com/enjoy-digital/litedram/issues/156, so you can review the tests when you have some time.
`test_bist_csr_cdc` fails as I described in https://github.com/enjoy-digital/litedram/pull/165#issuecomment-599629743.

enjoy-digital:
Thanks, this seems fine and we can probably merge with a green Travis-CI. Can you just:
- move the DMA specific tests to`test_dma.py`?
- comment the failing test and create an issue for that? (i could have a closer look at it).

enjoy-digital:
Thanks, merged.

PR #170, opened 2020-03-24
Add tests for litedram.frontend.adaptation

jedrzejboczar:
Part of  https://github.com/enjoy-digital/litedram/issues/156.

This adds tests for up/down converters with different conversion ratios and for CDC.

I've also added [test_converter_up_not_aligned](https://github.com/antmicro/litedram/blob/jboc/unit-tests/test/test_adaptation.py#L199) for up-converter when the pattern is not composed of groups of `ratio` consecutive words ([example pattern](https://github.com/antmicro/litedram/blob/0afacba2ca4f1b69831bdd1b83c4cd09875cd763/test/common.py#L356)), but this currently fails so I commented out that test. Do we want to support this case in `LiteDRAMNativeWritePortUpConverter` in the future or this test case is unnecessary?

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -5,6 +5,7 @@
 
 import os
 import random
+import itertools
 from operator import or_
 
 from migen import *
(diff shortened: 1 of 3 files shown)


jedrzejboczar:
I've additionally changed the tests that were commented out to use `@unittest.skip("...reason...")`, so that they will be visible during test runs.

enjoy-digital:
Thanks, this seems fine. We can merge and i'll have a closer look while validating checking that points of #156 are covered.

enjoy-digital:
I'll also have a closer look a the unit-test that is commented out to see if we should need to have it.

PR #172, opened 2020-03-26
modules: add MTA4ATF51264HZ DDR4 SO-DIMM

piotr-binkowski:
This PR adds DDR4 SO-DIMM module that works on the ZCU104 board
This module was tested as part of litex-hub/litex-boards#49


enjoy-digital:
Thanks, merged.

Issue #155, opened 2020-03-05
Add unit-tests to litedram.core modules


enjoy-digital:
In order to be able to refactor and rework the core modules of LiteDRAM, we first need to improve unit-tests of the core modules.

Some work has been done recently to improve the Refresher and add new features or verify the Timing Controller were optimal, while doing this work, unit-tests have been added: 
- https://github.com/enjoy-digital/litedram/blob/master/test/test_refresh.py
- https://github.com/enjoy-digital/litedram/blob/master/test/test_timing.py

 We should make sure all core modules have similar unit-tests:
- [x] litedram.common.tXXDController
- [x] litedram.common.tFAWController
- [x] litedram.core.bandwidth.Bandwidth
- [x] litedram.core.bankmachine.BankMachine
- [x] litedram.core.multiplexer._CommandChooser
- [x] litedram.core.multiplexer.__Steerer
- [x] litedram.core.multiplexer.Multiplexer
- [x] litedram.core.crossbar.LiteDRAMCrossbar
- [x] litedram.core.refresher.RefreshSequencer
- [x] litedram.core.refresher.Refresher

enjoy-digital:
Thanks @jedrzejboczar for the work, i've been able to review it and we can close this now.

PR #173, opened 2020-03-27
Add litedram.core.multiplexer tests

jedrzejboczar:
Part of https://github.com/enjoy-digital/litedram/issues/155.

This is still WIP, but I wanted to consult something.

I've started by adding `_CommandChooser` tests, but some of them fail and I wonder if I understand correctly what `_CommandChooser` is supposed to do. The current logic
https://github.com/enjoy-digital/litedram/blob/b06e946d09807f3ab9b2e72f9c599851ab8221b4/litedram/core/multiplexer.py#L43-L47
seems to be wrong in some cases, e.g.
* with `want_cmds=1` and `want_activates=0`, ACTIVATE commands will still be considered, because they have both `is_write=0` and `is_read=0` (and we have both `want_reads=0` and `want_writes=0`), so `read` and `write` are true, and thus `(read & write)` is true, even though `command` is false
* all `want_*` are 0, then both `read` and `write` are 1 for requests with `request.is_cmd == 1` so the command is considered valid
* with `want_writes=1` and `want_reads=1`, both write and read commands will be considered invalid (but we never want both at the same time, so it is probably a minor problem)

The following tests fail:
* `test_selects_cmds_without_act`
* `test_selects_nothing`
* `test_selects_nothing_when_want_activates_only`
* `test_selects_writes_and_reads`

I'll continue now with the tests for `_Steerer` and `Multiplexer`.


enjoy-digital:
Thanks, i won't probably be able to review it today, but will try to do it early next week. You can indeed continue on things that are not blocking your for now.

jedrzejboczar:
I've finished adding tests for `_Steerer` and `Multiplexer`.

This is a lot of code, but I tried to add a short description to each test, so that it is easier for you to check what functionality is being tested and tell me if I missed something important.

I also moved `_CommandChooser` and `_Steerer` tests to separate files, as the file has grown to 1068 lines.

The 4 tests mentioned in the PR description still fail, so the CI is red, but all other tests are ok. If you want I can skip these tests and create an issue for that.

enjoy-digital:
Thanks, sorry i haven't been able to review the previous code yet. You can indeed skip the failing tests and create an issue for that. I could look and help you getting this working or fix myself while reviewing.

I'm a bit lagging on the reviews, so if you think the Multiplexer if well covered you can move to another module and i'll add eventual comments on the Multiplexer tests later here or will merge and do the changes while doing the review.

jedrzejboczar:
Ok, I'll create an issue in a moment and will continue with other modules.

enjoy-digital:
@jedrzejboczar: thanks for the very good work, i did a first review and we can merge it.

PR #175, opened 2020-04-03
Add core.bandwidth tests

jedrzejboczar:
Requires https://github.com/enjoy-digital/litedram/pull/173
Part of https://github.com/enjoy-digital/litedram/issues/155

This PR adds tests of `Bandwidth`. It requires some changes from https://github.com/enjoy-digital/litedram/pull/173 but since that PR is ready to be merged, you can just review this new one after the merge, would it be ok?

Regarding this PR, I noticed that `Bandwidth` misses a command if it appears on period boundary (i.e. in the `If(period, ...)` condition in [bandwidth.py](https://github.com/enjoy-digital/litedram/blob/master/litedram/core/bandwidth.py)). I added a check for that and increased command counters width by 1 (as during N-cycle period we can have from 0 to N inclusive commands).


enjoy-digital:
@jedrzejboczar: thanks for the tests, great if you found a corner case and have been able to fix it with these.

PR #178, opened 2020-04-10
Add litedram.core.crossbar tests

jedrzejboczar:
Closes https://github.com/enjoy-digital/litedram/issues/155.

This PR adds core.crossbar tests of address mapping, arbitration, locking and different port types. It also includes a small fix of core.bankmachine tests.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -36,6 +36,58 @@ def timeout_generator(ticks):
     raise TimeoutError("Timeout after %d ticks" % ticks)
 
 
+class NativePortDriver:
+    """Generates sequences for reading/writing to LiteDRAMNativePort
+
+    The write/read versions with wait_data=False are a cheap way to perform
+    burst during which the port is being held locked, but this way all the
+    data is being lost (would require separate coroutine to handle data).
+    """
+    def __init__(self, port):
+        self.port = port
+
+    def read(self, address, wait_data=True):
+        yield self.port.cmd.valid.eq(1)
+        yield self.port.cmd.we.eq(0)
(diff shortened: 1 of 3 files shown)


jedrzejboczar:
Unit tests are ok, Travis failed later when downloading packages during simulations.

mithro:
@jedrzejboczar - You should be able to restart a failed travis build? Can you see if you can see it in the Travis UI?

jedrzejboczar:
Hmm, I can't see any option for that:
screenshot_2020-04-10_16-21-22
I've checked that I am signed in. Should it appear on the right as shown [here](https://stackoverflow.com/a/17624403/8385882)? I've tried to check it on a repo to which I have write access and I couldn't find it either. It is confusing, as I can remember that I was able to restart builds, at least for repos with write access, so I wonder what I am missing.

jedrzejboczar:
Ok, sorry for confusing you, I've checked in another browser and the button appears for a build on repo where I have write access. I cannot restart the build for this PR though, but I guess I can only do that for a failing one, am I correct?

mithro:
@jedrzejboczar -- You probably don't have the correct permissions on the LiteDRAM GitHub repository to restart builds it seems. As I can do that, @enjoy-digital can you give @jedrzejboczar the same level of permissions as me?

enjoy-digital:
@jedrzejboczar: sorry, i thought you already had the permissions, it should be fine now.

enjoy-digital:
Thanks for the great work @jedrzejboczar, i did a first review and merged it.

PR #181, opened 2020-04-15
Add option to load module data from DDR3 SPD EEPROM

jedrzejboczar:
This PR would adds option to load parameters of `SDRAMModule` from raw SPD data.
For now this is only implemented for DDR3 memory, but could be later extended for other types.

I've created some tests based on reference SPD contents from Micron's website.
I've also fixed `nrows` in `MT8KTF51264`.


jedrzejboczar:
Some modules' timings seem to be higher than in the datasheets, is there a reason for that? Should we update the timings?

For example, MT18KSF1G72HZ for speedgrade 1G4 (DDR3-1333) has `tRP = tRCD = 15` instead of 13.125 as in [its datasheet](https://www.micron.com/~/media/documents/products/data-sheet/modules/sodimm/ksf18c256_512_1gx72hz.pdf). The value of 13.125 is also present in its Micron's reference SPD data.

For this reason I did not include timing parameters comparisons in the unit tests yet - then most tests would fail. I compare only the geometry from SPD data/our module and I also check the timings directly with those from datasheet.

enjoy-digital:
@jedrzejboczar: We should also use this work to do a review of the current timings. While doing that, can you modify the values you think should be modified and i'll also review this? Thanks.

jedrzejboczar:
I've updated the timings based on the values extracted from SPD data, but I have some notes.

Firstly, for some modules I encountered inconsistencies between `tRCD` and `tRP` values in the SPD data, in the SO-DIMM module datasheet and in the datasheet of the chip used in the module. In the end, I was chosing the values from SPD data when there existed any SPD reference for the given module, but I am a bit confused here.

Secondly, I see that `tRRD` is specified in `_TechnologyTimings`, so we cannot specify different values for different speed grades. In the [datasheet](https://www.micron.com/-/media/client/global/Documents/Products/Data%20Sheet/DRAM/DDR3/4Gb_1_35V_DDR3L.pdf) it depends on the speed grade, so I understand that for this reason we were just using the highest possible value. On the other hand, in a SO-DIMM module's datasheet we have a list of possible chip speed grades, so I updated `tRRD` to the highest of these values.

enjoy-digital:
Thanks @jedrzejboczar, this looks good and will be very useful. It's merged and issues have been created for the remaining points: https://github.com/enjoy-digital/litedram/issues/184 and https://github.com/enjoy-digital/litedram/issues/185.

PR #189, opened 2020-04-29
Add support for DDR4 RDIMMs

daveshah1:
I would appreciate some feedback on the structure, it seems like perhaps is_rdimm and drive strengths should be a module property, but I can't find a nice way of glueing that in.

This also still needs regression testing on a non-RDIMM platform too, I can do that if you want.

### DIFF
--- a/litedram/common.py
+++ b/litedram/common.py
@@ -173,6 +173,7 @@ class PhySettings(Settings):
                  cl, read_latency, write_latency, nranks=1, cwl=None):
         self.set_attributes(locals())
         self.cwl = cl if cwl is None else cwl
+        self.is_rdimm = False
 
     # Optional DDR3/DDR4 electrical settings:
     # rtt_nom: Non-Writes on-die termination impedance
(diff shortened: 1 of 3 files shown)


daveshah1:
See also enjoy-digital/litex#490

enjoy-digital:
Thanks! i'll review it in the next days.

enjoy-digital:
In `modules.py`, we should probably create specialized SDRAMModules:
````
class SDRAMRegisteredModule(SDRAMModule): registered = True

class SDRModule(SDRAMModule): memtype = "SDR"
class SDRRegisteredModule(SDRAMRegisteredModule): memtype = "SDR"

...

class DDR4Module(SDRAMModule): memtype = "DDR4"
class DDR4RegisteredModule(SDRAMRegisteredModule): memtype = "DDR4"
````

And use them as the base class for the modules. This would simplify the definition and would provide the `registered` attribute that would be used by the PHY/Core.

In `PhySettings`, we could rename  `is_rdimm` to `registered` attribute. Does it make sense for you? If so, i'm happy to do the changes and do the regression testing on targets with Unbuffered memory. (unless you want to do it, but the improvements here are more related to LiteDRAM itself than your PR).

daveshah1:
Yes, that sounds good. The RCD drive strength settings come from the SPD data and so should also be a property of DDR4RegisteredModule. Please feel free to make these changes.

enjoy-digital:
Thanks, merged. I'll do the adaptations.

PR #190, opened 2020-04-29
WIP: Implement cmd_latency calibration 

jedrzejboczar:
This PR would introduce possibility to change `cmd_latency` dynamically during SDRAM operation, which can be used to find the correct value in the controller software. 

At this moment the implementation is only for USDDRPHY and breaks other PHYs. This implementation introduces CSRs for `cmd_latency`, `wrphase` and `wrcmdphase`. These are used in `Multiplexer` (`steerer_sel`) and in `LiteDRAMCrossbar` (`write_latency`). Could you please take a look at this to check if this is a good direction? I am somewhat worried that I had to add `Signal`s to `PhySettings`, it seems wrong considered that those are meant to be constant.

I have some corresponding changes to Litex (https://github.com/enjoy-digital/litex/compare/master...antmicro:jboc/sdram-calibration), but will create a PR when this one is ready.

### DIFF
--- a/litedram/core/crossbar.py
+++ b/litedram/core/crossbar.py
@@ -61,12 +61,15 @@ class LiteDRAMCrossbar(Module):
     def __init__(self, controller):
         self.controller = controller
 
-        self.rca_bits         = controller.address_width
-        self.nbanks           = controller.nbanks
-        self.nranks           = controller.nranks
-        self.cmd_buffer_depth = controller.settings.cmd_buffer_depth
-        self.read_latency     = controller.settings.phy.read_latency + 1
-        self.write_latency    = controller.settings.phy.write_latency + 1
+        self.rca_bits          = controller.address_width
+        self.nbanks            = controller.nbanks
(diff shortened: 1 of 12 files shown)


daveshah1:
For the Alveo board, I've recently found that different command latencies for bits 31..0 and 63.32 will probably be needed. Do you think this is something that might be possible to add to this?

enjoy-digital:
Thanks @jedrzejboczar, i'll look at it. @daveshah1: We'll think about this for the Alveo case yes. In your case, the write leveling is also not optimal and for now we are setting a common delay to the commands/clock for all modules but we could use individual delay for each module, which would probably help for the Alveo. So that's possible that this PR + individual delays on the commands/clock will improve the situation on the Alveo. If not, then we could try to support individual cm_latency.

enjoy-digital:
Thanks @jedrzejboczar, the issue we had behind that was in fact that write latency calibration was missing and i added it recently. Supporting dynamic `cmd_latency` should no longer be needed.

Issue #182, opened 2020-04-23
Refresh S6DDRPHY


enjoy-digital:
The Spartan6 S6DDRPHY needs to be refreshed with the simplifications/improvements made to the others PHYs.

cklarhorst:
@enjoy-digital 
Could you please elaborate a bit more on what simplifications/improvements are not implemented/missing for S6DDRPHY?
Is one of the missing features that S6DDRPHY doesn't use IDELAY/ODELAY for data alignment?

enjoy-digital:
@cklarhorst: sorry i forgot to answer, the S6DDRPHY is the first PHY that has been created and we improved things since then, so the architecture would need to be updated a bit and it's also possible some features are missing (DQS Pre/Post amble for example). That's not necessarilly something easy to work on when not familiar with LiteDRAM, so i was planning to work on that, but if you want to have a closer look, i could help you getting started.

norbertthiel:
@enjoy-digital 
I am currently implementing Litex on a custom XC6SLX16-3 board which comes with LPDDR (Winbond W949D6KBH)
It works fine using parameters copy from the pipistrello board - CPU clock at 83.3 MHz and memory at 166MHz
* trying 200MHz memory clock and 100MHz CPU clock results in mem_test failing 

So, before start digging into this, I am wondering if you started refreshing the S6DDRPHY already?

enjoy-digital:
@norbertthiel: Hi, the work has not been initiated in this no.

norbertthiel:
> @norbertthiel: Hi, the work has not been initiated in this no.

Hi and thanks for the reply. Any short-term plans for initiating this?

Maybe you can point me were to start looking at to fix my "problem":
* 83,3MHz sys, 166MHz sdram_half, 333MHz sdram_full (pippistrello org settings)  => works
* 90Mhz sys, 180Mhz sdram_half, 360MHz sdram_full => when using "mem_write 0x40000000 0x01020304 8" I get all 8 values written = but at an offset of roughly 124 bytes there is a garbage word written to memory (3 garbage words written at offset 0x8e, 0x9e, 0x116) ... I have no idea were this is coming from  - any idea/hint?

memory supports 200MHz, FPGA has speedgrade 3

norbertthiel:
@enjoy-digital 
like this:
<img width="721" alt="image" src="https://user-images.githubusercontent.com/44259856/107612809-5b968100-6c47-11eb-8427-001991b4edd2.png">

[...]

PR #196, opened 2020-05-21
Integrate micron memory model simulation

mglb:
Issues:
* PHY needs 4 phases to use BL=8. BL for DDR3 is changed to 4 for now.
* Invalid DQS during writing
* Xilinx primitives (especially OSERDES/ISERDES) used in the PHY must be replaced

Before use:
* Download verilog model from https://www.micron.com/products/dram/ddr3-sdram/part-catalog/mt41k512m16ha-125
* Unpack it to ./thirdparty/micron
* Run `./patch-micron.sh && ./patch-xil-models.sh` in thirdparty directory

Use with https://github.com/enjoy-digital/litex/pull/536

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "thirdparty/xil-models"]
 	path = thirdparty/xil-models
-	url = git@github.com:dirjud/Nitro-Parts-lib-Xilinx.git
+	url = https://github.com/dirjud/Nitro-Parts-lib-Xilinx.git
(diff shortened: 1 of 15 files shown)


enjoy-digital:
Thanks @mglb for the initial work, there are several changes in the PR and i need to think a bit about it. I'm also working on a bench for LiteDRAM with automated tests and i'm wondering if this should be part of it. Having a GenericDDR3PHY is a good idea, but this is not that simple in practice since it's not easy to have a generic API around the IO primitives for the modes used in the DRAM PHYs, so we also need to think about it. We should also probably work on https://github.com/enjoy-digital/litedram/issues/182 first before trying to do a GenericDDR3PHY.

enjoy-digital:
Sorry but i just described in enjoy-digital/litex#536 i think this should be maintained separately.

Icenowy:
For GenericDDR3PHY, try something like https://github.com/ZiyangYE/General-Slow-DDR3-Interface ? (Although this requires the DRAM chip that implements DLL-off mode)

PR #197, opened 2020-05-22
Allow generation of a standalone sim model

ozbenh:
(This depends on the corresponding branch in LiteX)

This allows litedram_gen --sim to be used to generate a simulated litedram model from
the same .yml file as a real synth model.


enjoy-digital:
Thanks, this has been merged and simplified a bit with https://github.com/enjoy-digital/litedram/commit/d1db115d6c63c943fd5ca20914a642d670795f0c. The changes in LiteX are not yet merged but i'll look at it soon.

Issue #194, opened 2020-05-15
ECP5: Use ODDRX1F for address and command bus


kingoflolz:
It should be possible to use ODDRX1F instead of ODDRX2F for the address and command bus, so that these signals can be placed on the top IO bank instead of just the left and right.

This seems like a relatively straightforward change, and I would be happy to provide a pull request, unless there are some nuances I'm not seeing?

enjoy-digital:
Hi @kingoflolz, if you do the change and confirm the Versa ECP5 target (https://github.com/enjoy-digital/litex/blob/master/litex/boards/targets/versa_ecp5.py) is still building fine with it, that's probably something we could merge yes.

enjoy-digital:
@kingoflolz: sorry i won't be able to spend time investigating this and i'm not sure it will be the case for someone else in the near future. I'll however be happy to test and integrate a validated solution if provided.

madscientist159:
@enjoy-digital Since it's been a year, and we've run straight into this problem on some custom hardware, any chance of re-visiting this request?

enjoy-digital:
@madscientist159: If I remember correctly replacing the ODDRX2F with ODDRX1F was not as straightforward as expected and some time would need to be spent understanding why. If we want to go further on this, I would happy to test and do the integration of a validated solution (or spend time on this is someone has funds for it).

madscientist159:
We've investigated the issue and come up with a working solution here, tested to work on the Versa board:
https://gitlab.raptorengineering.com/kestrel-collaboration/kestrel-litex/litedram/-/tree/ecp5ddr1x

The problem wasn't clock skew between `ECLK`/`SCLK`, it was the phase shift introduced by the `ODDRX2` primitive configured the way it was in LiteX.  The `ODDRX1` primitive has less effective latency when clocked from `SCLK`, and its output needed to be delayed by one `ECLK` period as a result.

It isn't particularly easy to generate a low-skew inverted clock signal (needed to drive the `ODDRX1` primitive) from an existing edge clock on these devices.  We settled on a second `CLKDIVF` with some logic to use the `ALIGNWD` signal to ensure the clock is inverted, but the obvious disadvantage of this approach is the need to allow the signals to cross to the other side of the device (wasting `ECLK` resources).

I do note that Lattice's documentation encourages the use of the TOP I/O bank for address and command lines.  As a result I think it is important that the option be available to use the device as Lattice specifies, even if the logic for that case is a bit more complex.

[...]

PR #198, opened 2020-05-23
use ODDRX1F for command and address output on ECP5

kingoflolz:
From #194 

Confirmed that versa_ecp5 compiles, but not tested on real hardware

### DIFF
--- a/litedram/phy/ecp5ddrphy.py
+++ b/litedram/phy/ecp5ddrphy.py
@@ -162,25 +162,19 @@ class ECP5DDRPHY(Module, AutoCSR):
 
             # Addresses and Commands ---------------------------------------------------------------
             for i in range(addressbits):
-                self.specials += Instance("ODDRX2F",
+                self.specials += Instance("ODDRX1F",
                     i_RST  = ResetSignal("sys2x"),
-                    i_ECLK = ClockSignal("sys2x"),
                     i_SCLK = ClockSignal(),
                     i_D0   = dfi.phases[0].address[i],
-                    i_D1   = dfi.phases[0].address[i],
-                    i_D2   = dfi.phases[1].address[i],
-                    i_D3   = dfi.phases[1].address[i],
+                    i_D1   = dfi.phases[1].address[i],
                     o_Q    = pads.a[i]
                 )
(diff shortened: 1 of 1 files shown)


daveshah1:
The latency between the two primitive may well be different, so a test on hardware would definitely be good.

enjoy-digital:
@kingoflolz: thanks for #198, i did a test with it on Trellisboard but it's not working unfortunately. This would need to be investigated further in simulation or/and on hardware.

kingoflolz:
Could the issue be with using the sys2x reset with the sys clock domain on the ODDRX1F? I do not have the equipment necessary to really dig into this (on hardware) unfortunately.

daveshah1:
I think the bigger problem is likely the unknown phase shift between sys2x (which is creating the output clock for the DDR) and sys (which is now clocking the command/address outputs). It might be possible to calibrate out this delay with an output DELAYF on the command/address signals but this feels like something that should only be enabled if really needed and ODDRX2F used for all platforms with non-broken PCB layouts.

enjoy-digital:
@kingoflolz: do you have a specific use case in mind for this? Because since this does not seem to be as simple as expected, like @daveshah1, i think this is something we should only be supporting if really required and if a working solution is provided.

kingoflolz:
I'm doing bringup on a custom ECP5 board which has some command and address pins routed to the top IO bank. As it lookes like its going to be quite complicated, I might just give up dram for this revision and route everything to the left and right IO bank for the next revision.

kingoflolz:
I do find it interesting that the lattice datasheet does suggest the top IO bank can be used for DDR, I wonder if the lattice DDR3 IP handles that case.
image

enjoy-digital:
@kingoflolz: thanks for the info. That's indeed probably possible, but requires some investigation to understand what is miss-behaving. I'll try to see if i can spend 1-2 hours on this but not sure i'll be able to investigate more.

enjoy-digital:
@kingoflolz: sorry i won't be able to spend time investigating this and i'm not sure it will be the case for someone else in the near future. I'll however be happy to test and integrate a validated solution if provided.

PR #199, opened 2020-05-27
modules: add function for parsing SPD EEPROM dumps from BIOS firmware

jedrzejboczar:
This adds a function for parsing SPD EEPROM data in the form dumped in BIOS.


enjoy-digital:
Thanks, merged.

PR #201, opened 2020-06-01
modules/spd: save SPD data in SDRAMModule

jedrzejboczar:
Makes `SDRAMModule` store `spd_data` as a class member when generating from SPD data. This can be then used to add ROM with the data and verify it at runtime in BIOS (requries LiteX PR).

This also includes some minor changes to `DDR3SPDData` parser that I made while implementing `DDR4SPDData` parser. I didn't include DDR4 parser, as I still had some errors when using it. I will add it in a separate PR when I find what was the exact problem.

### DIFF
--- a/litedram/modules.py
+++ b/litedram/modules.py
@@ -48,46 +48,49 @@ def _twos_complement(value, nbits):
 def _word(msb, lsb):
     return (msb << 8) | lsb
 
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks, merged.

Issue #200, opened 2020-05-28
Memtest fails on ZCU104 board


kowalewskijan:
I used the latest code to build Litex for ZCU104 board but in the result I got a following error during `memtest`:
```
--========== Initialization ============--
Initializing SDRAM...
SDRAM now under software control
Write leveling:
Command/Clk scan:
|00000000|0000|0000|0000| best: -1
Data scan:
m0: |1111111111111111111111| delay: -1
m1: |1110000000000000000001| delay: -1
m2: |1111111111111111111111| delay: -1
m3: |1111100000000000000000| delay: -1
m4: |1111111111111111111111| delay: -1
m5: |1111111111111111111111| delay: -1
m6: |1111111111000000000000| delay: -1
m7: |1111111111111111111111| delay: -1
Read leveling:
m0, b00: |00000000000000000000000000000000| delays: -
m0, b01: |00000000000000000000000000000000| delays: -
m0, b02: |00000000000000000000000000000000| delays: -
m0, b03: |00000000000000000000000000000000| delays: -
m0, b04: |00000000000000000000000000000000| delays: -
m0, b05: |00000000000000000000000000000000| delays: -
m0, b06: |00000000000000000000000000000000| delays: -
m0, b07: |00000000000000000000000000000000| delays: -
m0, b08: |00000000000000000000000000000000| delays: -
m0, b09: |00000000000000000000000000000000| delays: -
m0, b10: |00000000000000000000000000000000| delays: -
m0, b11: |00000000000000000000000000000000| delays: -
m0, b12: |00000000000000000000000000000000| delays: -
m0, b13: |00000000000000000000000000000000| delays: -
m0, b14: |00000000000000000000000000000000| delays: -
m0, b15: |00000000000000000000000000000000| delays: -
best: m0, b00 delays: -
m1, b00: |00000000000000000000000000000000| delays: -
m1, b01: |00000000000000000000000000000000| delays: -
m1, b02: |00000000000000000000000000000000| delays: -
m1, b03: |00000000000000000000000000000000| delays: -
m1, b04: |00000000000000000000000000000000| delays: -
m1, b05: |00000000000000000000000000000000| delays: -
m1, b06: |00000000000000000000000000000000| delays: -
m1, b07: |00000000000000000000000000000000| delays: -
m1, b08: |00000000000000000000000000000000| delays: -
m1, b09: |00000000000000000000000000000000| delays: -
m1, b10: |00000000000000000000000000000000| delays: -
[...]

PR #204, opened 2020-06-02
Add DDR4 SPD EEPROM data parser

jedrzejboczar:
This adds a parser for data from DDR4 SPD EEPROM. It uses `tXX_L` timings as we do not support `tXX_S` yet. It also checks page size to determine minimal `tFAW` in clocks.

I've tested this on ZCU104 with MTA4ATF51264HZ and the timings taken from SPD are correct (I also verified SPD data with reference on Micron's website), but when I was using the timings taken from SPD the module failed during leveling. 

I checked that this is not due to the generation process, because changing the timings in the module definition gave the same results. I found that the only timings that differed enough to result in different `timing_settings` were `tFAW` (5 -> 7) and `tCCD` (1 -> 2). I wonder why this would result in leveling failing, when the timings increased, not decreased. Could this somehow influence leveling process?

I added SPD parsing tests in `test_modules.py` but one is being skipped for now. Maybe we can merge this and I can create an issue with information about this problem?


enjoy-digital:
Thanks @jedrzejboczar, i will have a closer look, but just for info, it seems there is a regression on the ZCU104: https://github.com/enjoy-digital/litedram/issues/200 I will look at this, but this could explain your issue. Have you tested with the default SDRAMModule?.

jedrzejboczar:
@enjoy-digital I was already testing with the change described in https://github.com/enjoy-digital/litedram/issues/200 that seems to fix the problem. Leveling stopped working only after changing the timings.

enjoy-digital:
Thanks @jedrzejboczar, the parser support itself seems fine, so we can indeed merge it. For the issue with on hardware, can you create an issue for it as you were suggesting? I could have a closer look and could also do some tests on other boards. The leveling is slow process so `tFAW` and `tCCD` should not be impacting it but we will need to investigate.

Issue #203, opened 2020-06-02
Improve LiteDRAMFIFO behaviour on an interrupted stream.


enjoy-digital:
The current implementation of the FIFO expects a continous stream of data. When the stream is interrupted, due to the `read_threshold` some data that should be emitted on the `source` are blocked in the FIFO until new data are written to the `sink`. We should implement a mechanism to avoid this and emit the data. The `read_threshold` was implemented to handle write-to-read latency and be sure write was done to the DRAM before initiating the read, but we can probably use internal signals of the controller to increment the level on a write exactly when the write burst is done to the DRAM and avoid the `read_threshold`.

PR #205, opened 2020-06-03
frontend/fifo: increase FIFO level only after data has actually been written

jedrzejboczar:
Closes https://github.com/enjoy-digital/litedram/issues/203

This removes the need for `read_threshold` by using `port.wdata.ready` (or `port.w.ready` for `LiteDRAMAXIPort`) to increment FIFO level. This way data is available for read only after it has been actually written to memory. The tests have been updated and `write_threshold` has also been removed as now it is not needed.


enjoy-digital:
Thanks, this makes sense. Merged.

PR #210, opened 2020-07-26
Add support for TDQS mode.

oskirby:
This adds an optional argument to `add_electrical_settings()` that can enable the TDQS mode for x8 DDR3 modules. Normally this would be used when mixing x4 and x8 modules (eg: heterogenous DIMMs), but it can also serve as a workaround boards with missing or broken DM signals.

I had to add this as a workaround for the rev0 Logicbone prototypes because the DM signals on the DDR3 modules wound up floating due to a combination of layout errors. The downside of enabling TDQS mode is that the data mask signals no longer have any effect, and you can only ever write a full burst.

Given that this is a workaround for a prototype board that may not ever see the light of day, it's also okay to reject this PR.

### DIFF
--- a/litedram/common.py
+++ b/litedram/common.py
@@ -179,7 +179,8 @@ class PhySettings(Settings):
     # rtt_nom: Non-Writes on-die termination impedance
     # rtt_wr: Writes on-die termination impedance
     # ron: Output driver impedance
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @oskirby, this could be useful for other debug cases so happy to merge it.

PR #217, opened 2020-10-01
Update .travis.yml

pragyabansal02:


enjoy-digital:
@pragyabansal02 thanks for the PR, but `AS4C267M27D4A` is not a module we are supporting. If that's a module you added, can you also contribute it?

PR #222, opened 2020-10-13
init: Cast DDR4 RCD fine_speed to int

daveshah1:
Fixes `TypeError: unsupported operand type(s) for |: 'int' and 'float'`
for some clock frequencies.


enjoy-digital:
Thanks.

PR #223, opened 2020-10-20
ecp5ddrphy: Fix DELAYF initial value

gregdavill:
I've seen intermittent failures with recent builds. I suspect it is related to the DELAYF primitive not being correctly loaded with an initial value. Holding LOADN LOW ensures a value is correctly loaded.

I'm not exactly sure why we are only seeing this issue now. But I believe this is similar behaviour to what I was seeing with the DELAYF primitives in the liteeth ecp5 phy.


enjoy-digital:
Thanks, i also had this behavior but the fact it's not deterministic make it difficult to track it down. I'll also do more testing with you change applied.

PR #226, opened 2020-12-31
modules: add MT48LC32M8 SDR module

mdpye:
The 8 meg x 8 x 4 bank variant of the existing MT48 series modules.

The chip is used in the SDRAM and HDMI add-on shields for the Alchitry MojoV3 hobbyist platform: https://alchitry.com/products/mojo-v3


enjoy-digital:
Thanks @mdpye, that's merged.

PR #227, opened 2021-01-22
phy/ecp5ddrphy: invert the clock signal if necessary.

garytwong:
(Please note this pull request depends on https://github.com/enjoy-digital/litex/pull/781.)

### DIFF
--- a/litedram/phy/ecp5ddrphy.py
+++ b/litedram/phy/ecp5ddrphy.py
@@ -188,7 +188,7 @@ class ECP5DDRPHY(Module, AutoCSR):
                     i_RST  = ResetSignal("sys"),
                     i_SCLK = ClockSignal("sys"),
                     i_ECLK = ClockSignal("sys2x"),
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, is it due to a clk pin inversion on a custom board?

garytwong:
Yes indeed... hard to imagine somebody would connect + to - and - to +.  But if that happens, then it's easier to fix the FPGA config than the PCB :smile:

enjoy-digital:
Thanks for the information. The `Inverted` attribute has been done as a quick workaround and not necessarily something I would like to keep (or at least encourage the use) until we found a better solution. So I added with https://github.com/enjoy-digital/litedram/commit/5a114be7e597767b4540a5292e3f13583f4624e3 a `clk_polarity` parameter to ECP5DDRPHY that you can use on you board to invert `clk` polarity.

garytwong:
Thanks! The `clk_polarity` parameter works fine.

PR #230, opened 2021-01-29
test: improve error messages when comparing files in test_init.py

jedrzejboczar:
This is a minor change, but I hope it will improve readability of errors in `test_init.py`. It will now print a diff when the files do not match. It also avoids creating the temporary file (only reads the reference file).


enjoy-digital:
Thanks @jedrzejboczar, this will indeed probably be convenient, that's merged.

PR #224, opened 2020-12-02
[WIP] Add LPDDR4 PHY

jedrzejboczar:
This would add support for LPDDR4 PHYs in LiteDRAM. It is still WIP but will be useful to track progress.

In the current implementation there is LPDDR4PHY which will be a base class for a PHY for series 7 which will yet have to be added. Base PHY's job is to convert DFI commands to LPDDR4 commands format and provide signals that are ready to be serialized using hardware blocks in a concrete PHY. 

LPDDR4 sends commands on an CA[5:0] SDR  lines. DRAM commands consist of 1 or 2 "subcommands" (e.g. ACT = ACT-1 + ACT-2) and each subcommand is sent over 2 clock cycles. Currently the translation has been implemented and there are tests in Migen simulator that verify serialization of CA. Due to 16n prefetch of LPDDR4 the PHY uses 8 DFI phases to avoid complications if the write/read bursts would have to be sent over more than 1 cycle of DRAM controller clock.

PR #231, opened 2021-01-29
Allow to pass all module timings in the format (ck, ns)

jedrzejboczar:
This is a part of splitting https://github.com/enjoy-digital/litedram/pull/224 into smaller PRs.

This PR makes it possible to pass all timings as `(ck, ns)` tuples (here `ns` can be `None`) or just `ns` float.

### DIFF
--- a/litedram/modules.py
+++ b/litedram/modules.py
@@ -294,6 +294,12 @@ def parse_spd_hexdump(filename):
 
 # SDRAMModule --------------------------------------------------------------------------------------
 
+Frac = namedtuple("Frac", ["num", "denom"])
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @jedrzejboczar for splitting #224 in smaller PRs, this looks good (the others PRs too). I'll have a closer look and test/merge in the next days.

enjoy-digital (approved):
Thanks @jedrzejboczar, this looks good. I also did some comparisons of the generated verilog before and after applying this PR on various targets and the timings are computed identically.

PR #232, opened 2021-01-29
init: make the write leveling MR bit configurable

jedrzejboczar:
This is a part of splitting https://github.com/enjoy-digital/litedram/pull/224 into smaller PRs.
Requires corresponding PR in litex: https://github.com/enjoy-digital/litex/pull/791.

In LPDDR4 the write leveling mode is configured with MR2[7] instead of MR1[7]. This PR changes the way `get_sdram_phy_init_sequence` returns mode register values. Now it should return a dictionary `{mr_num: mr_value, ...}`. Instead of generating `DDRX_MR1` define we now have 3 `DDRX_MR_WRLVL_*` defines that specify the mode register changes required to enter/leave write leveling mode.

Note that `DDRX_MR_WRLVL_BIT` is the bit that gets flipped (with xor), so `DDRX_MR_WRLVL_RESET` is the state when write leveling is disabled and software just filps given bit when entering write leveling mode and writes the original `DDRX_MR_WRLVL_RESET` value when leaving.

### DIFF
--- a/litedram/init.py
+++ b/litedram/init.py
@@ -220,7 +220,7 @@ def get_ddr3_phy_init_sequence(phy_settings, timing_settings):
         ("ZQ Calibration", 0x0400, 0, "DFII_COMMAND_WE|DFII_COMMAND_CS", 200),
     ]
 
-    return init_sequence, mr1
+    return init_sequence, {1: mr1}
 
 # DDR4 ---------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


jedrzejboczar:
`test_genesys2` fails because the litex used in CI doesn't include https://github.com/enjoy-digital/litex/pull/791

enjoy-digital (approved):
Thanks @jedrzejboczar, this makes sense and looks good. Initial return of mr1 was indeed a  quick workaround to get the confguration value of MR1 in the BIOS, thanks for making this more flexible.

PR #234, opened 2021-02-21
Add module name CLI option

craigjb:
Adds ability to specify a custom build name (Verilog module name) on the CLI. When using litedram in standalone mode and generating multiple DRAM interfaces, each generated litedram module needs to have a unique name otherwise they will conflict in the Verilog namespace.


enjoy-digital (approved):
Thanks @craigjb, I'm going to merge it but will just rename the args to --name.

PR #236, opened 2021-03-22
modules.py: Add MT41J256M8 (Passes mem_test on HW @ sys4x=500MHz, vex…

jersey99:
…riscv, marble board)

Adding a DDR3 SO-DIMM based module.


enjoy-digital:
Thanks @jersey99, this is merged.

PR #238, opened 2021-03-24
core/refresher: use A10=1 for an all-banks REF

jedrzejboczar:
Related to https://github.com/enjoy-digital/litedram/pull/224.

In LPDDR2+ and DDR5 the refresh command has two variants: per-bank and all-banks. This PR changes Refresher to always issue the all-banks variant. This should not affect other memory types, as they always refresh all banks, ignoring the address (it just must be a defined voltage level).


enjoy-digital (approved):
Thanks @jedrzejboczar, this makes sense.

PR #241, opened 2021-03-31
Fix Python header generation for LPDDR4

jedrzejboczar:
With the previous code generated lines included Python variable names instead of integers, e.g.
```python
("ZQ Calibration start", MPC.ZQC_START, SpecialCmd.MPC, dfii_command_we|dfii_command_cs, 200),
```
which failed as these names were not being imported.


enjoy-digital:
Thanks.

PR #245, opened 2021-04-22
add support for Winbond W9825G6KH-6 at 50MHz 1:2 rate

hansfbaier:
I finally came to the conclusion that although it only works well at 50MHz and 1:2 rate,
it still will be useful if that is documented.
I just realize many of the QMTech boards have the same chip, so having
it here will be good.

### DIFF
--- a/litedram/modules.py
+++ b/litedram/modules.py
@@ -535,6 +535,46 @@ class M12L16161A(SDRModule):
     technology_timings = _TechnologyTimings(tREFI=64e6/4096, tWTR=(2, None), tCCD=(1, None), tRRD=(None, 10))
     speedgrade_timings = {"default": _SpeedgradeTimings(tRP=15, tRCD=15, tWR=15, tRFC=(None, 55), tFAW=None, tRAS=40)}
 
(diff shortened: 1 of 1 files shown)


hansfbaier:
This is peculiar: The QMTech EP4CE15 board uses the above chip, see this 
[link](https://www.aliexpress.com/item/32949281189.html?spm=a2g0o.productlist.0.0.3ac9239dlOwckX&algo_pvid=77ad1d43-991c-4a7c-b896-356b1dbf9628&algo_expid=77ad1d43-991c-4a7c-b896-356b1dbf9628-0&btsid=0b0a555b16191314599938431e3172&ws_ab_test=searchweb0_0,searchweb201602_,searchweb201603_).
But according to litex-boards: qmtech_ep4ce15.py IS42S16160 should work.
I have to try that for the MiSTer board. Maybe we can then just alias the above chip for IS42S16160 or use the values of that one.

hansfbaier:
I tried the IS42S16160 and it actually works for the Winbond chip in 1:2@50MHz
But not for single rate and 50MHz/80MHz. Neither double rate@60MHz.

PR #247, opened 2021-04-27
Fix UpConverter reversed write mask

andrewb1999:
This pull request reverses the write mask bits in the UpConverter when reverse is set to True. Without this change `wdata_converter.source.we & wdata_sel` will result in a bad write mask as wdata_converter.source.we is reversed correctly but wdata_sel is not.

Signed-off-by: Andrew Butt <andrewb1999@gmail.com>


enjoy-digital:
Thanks @andrewb1999 for catching it, this is merged.

PR #248, opened 2021-04-28
init/lpddr4: make some settings configurable via phy_settings

jedrzejboczar:
Sometimes initialization parameters may need to be modified to tune the DRAM performance. This change allows to pass some key settings via `phy_settings`, the same way it is done for DDR3/DDR4.


enjoy-digital:
Thanks @jedrzejboczar, this is merged.

PR #250, opened 2021-05-01
Only update UpConverter sel when input valid

andrewb1999:
In the case where the input port valid signal goes low while the UpConverter is in the FILL stage, sel would continue to update. This would cause the fsm to prematurely leave the FILL stage.

Signed-off-by: Andrew Butt <andrewb1999@gmail.com>


enjoy-digital:
Thanks for catching this @andrewb1999, this is merged.

PR #252, opened 2021-05-17
init: generate sdram_phy.h in a way that allows to include it in multiple units

jedrzejboczar:
Currently `generated/sdram_phy.h` includes `sdram_dfii_pix_*data_addr` as variables (arrays). This prevents the header from being included in more than one .c file, because this results in multiple definitions of the variable. This PR replaces these variables using inline functions which look like this:
```c
static inline unsigned long sdram_dfii_pix_wrdata_addr(int phase){
	switch (phase) {
		case 0: return CSR_SDRAM_DFII_PI0_WRDATA_ADDR;
		case 1: return CSR_SDRAM_DFII_PI1_WRDATA_ADDR;
		case 2: return CSR_SDRAM_DFII_PI2_WRDATA_ADDR;
		case 3: return CSR_SDRAM_DFII_PI3_WRDATA_ADDR;
		case 4: return CSR_SDRAM_DFII_PI4_WRDATA_ADDR;
		case 5: return CSR_SDRAM_DFII_PI5_WRDATA_ADDR;
		case 6: return CSR_SDRAM_DFII_PI6_WRDATA_ADDR;
		case 7: return CSR_SDRAM_DFII_PI7_WRDATA_ADDR;
		default: return 0;
	}
}
```
This requires replacing previous access in form `sdram_dfii_pix_wrdata_addr[1]` with `sdram_dfii_pix_wrdata_addr(1)`.

Inlining many calls like this can potentially increase binary size. It would probably be best to just generate a matching .c file with these variables and have `extern ...` in the header. But currently Litex generates only header files so this is a quick workaround.

### DIFF
--- a/litedram/init.py
+++ b/litedram/init.py
@@ -735,7 +735,7 @@ def get_sdram_phy_c_header(phy_settings, timing_settings):
     # Commands functions
     for n in range(nphases):
         r += """
-__attribute__((unused)) static void command_p{n}(int cmd)
+__attribute__((unused)) static inline void command_p{n}(int cmd)
 {{
     sdram_dfii_pi{n}_command_write(cmd);
     sdram_dfii_pi{n}_command_issue_write(1);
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @jedrzejboczar, I also had this limitation in the past. The extra binary size should be reasonable, so this workaround is OK for now.

PR #253, opened 2021-05-26
fix gcc warning: function used but not defined

gsomlo:
The 'cdelay()' function is defined in LiteX 'liblitedram', so we
should not be using 'static' to declare it before calling it from
other C source files.

### DIFF
--- a/litedram/init.py
+++ b/litedram/init.py
@@ -730,7 +730,7 @@ def get_sdram_phy_c_header(phy_settings, timing_settings):
 
     r += "\n"
 
-    r += "static void cdelay(int i);\n"
+    r += "void cdelay(int i);\n"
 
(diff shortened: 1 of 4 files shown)


gsomlo:
CI *should* stop failing once https://github.com/enjoy-digital/litex/pull/923 is applied in LiteX, so maybe apply that one *first*.

gsomlo:
CI still fails, but only because it still pulls in a pre-923 commit of LiteX:

https://github.com/enjoy-digital/litedram/runs/2677064001#step:8:825

@enjoy-digital, please have a look and let me know what you think.

enjoy-digital:
Thanks @gsomlo, I merged it and will fix it if still failing.

PR #271, opened 2021-08-09
litedram_gen: Fix error with --sim option

antonblanchard:
It looks like commit 317072a1982d ("litedram_gen: Add initial SDRAM
support (with ULX3S example)") broke building with the --sim option.


enjoy-digital:
Thanks @antonblanchard, that's merged.

PR #273, opened 2021-09-01
Add RPC DRAM support

jedrzejboczar:
This PR adds a PHY for Etron RPC DRAM. The code directory structure resembles the one used for LPDDR4/LPDDR5.  The RPC PHY was actually the first of these PHYs and there are some design choices that could potentially be revised, applying the modifications from LPDDR4/5. Currently the PHY has some limitations, but it is functional and we were able to successfully use it on a modified Arty A7 board, although we were not able to reach the default Arty frequency due to hardware limitations. The current state of the PHY is described in more details in the readme under `litedram/phy/rpc/README.md`.


enjoy-digital:
This is merged, thanks a lot.

PR #274, opened 2021-09-09
Add AS4C128M16 DDR3L-1600 ram

teknoman117:
Add the AS4C128M16 DDR3L-1600 ram used on the Alchitry Au(+)

The tRFC timing is slightly worse than the MT41J128M16 which is already in litedram

Datasheet: https://www.alliancememory.com/wp-content/uploads/pdf/ddr3/Alliance%20Memory_2G%20128Mx16%20AS4C128M16D3LB-12BCN%20v1.0%20March%202016.pdf


enjoy-digital:
Thanks @teknoman117, that's merged.

PR #275, opened 2021-09-16
modules: add more RDIMM modules

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

This PR adds some additional modules for the RDIMM DDR4 type, from Micron and Hynix manufacturers.


acomodi:
@enjoy-digital PTAL

One question I have is how/where to get the `tWR` timing specification, as I couldn't find a reference to those in the data sheets.

kgugala:
I suppose we can read that from SPD

enjoy-digital:
@acomodi: Thanks this looks good and is merged. I'll have a look for `tWR`.

PR #278, opened 2021-09-22
Fix generation with no CPU

ozbenh:
The various UART bits in there need to be skipped

Signed-off-by: Benjamin Herrenschmidt <benh@kernel.crashing.org>


ozbenh:
I admit I don't really see why the tests are failing here ... Any idea ?

enjoy-digital:
Thanks @ozbenh, this is merged.

PR #279, opened 2021-09-24
Upconverter rewrite

mtdudek:
This series of commits improve frontend adapter bandwidth.

### DIFF
--- a/litedram/core/crossbar.py
+++ b/litedram/core/crossbar.py
@@ -76,7 +76,8 @@ class LiteDRAMCrossbar(Module):
 
         self.masters = []
 
-    def get_port(self, mode="both", data_width=None, clock_domain="sys", reverse=False):
+    def get_port(self, mode="both", data_width=None, clock_domain="sys", reverse=False,
+            rx_buffer_depth=4, tx_buffer_depth=4, cmd_buffer_depth=4, priority=0):
         if self.finalized:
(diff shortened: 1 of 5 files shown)


mtdudek:
@enjoy-digital what do you think?

enjoy-digital:
Thanks @mtdudek, I'm not able to review it now but will try to do so in the next days.

mtdudek:
So it failed on picolibc, and not on this upconverter, it states that meson is not installed

enjoy-digital:
Thanks @mtdudek, I've been able to have a first look. The new code is becoming really complex to follow and I'm not sure it will be possible to maintain it, so I'm not able to merge it now and will have to spend time understanding the previous limitations.

The primary issue is in fact coming from LiteVideo that should not require the use of an UpConverter and should directly handle access to the DRAM port with native data-width. This allows using a very simple specific UpConverter vs the generic (but complex) UpConverter here. The new VideoFramebuffer (now recommended vs LiteVideo VideoOut core) does not have this limitation and directly runs with native data-width: https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/video.py#L618.

enjoy-digital:
Thanks @mtdudek, another simpler solution has been implemented in: https://github.com/enjoy-digital/litedram/pull/385.

PR #280, opened 2021-10-03
Fix a few issues with ECP5 standalone generator

antonblanchard:


antonblanchard:
I haven't been able to get it working on ECP5 yet (so can't test), but these seem to be issues.

PR #284, opened 2021-11-06
W9825G6KH6 seems to use 8192 refresh cycles not 8000

hansfbaier:
image
I am not sure what they mean by 8K refresh cycles.
I assumed it is 8000 not 8192, but the datasheet does not make it clear anywhere else.
This part suggests it is 8192:
image


enjoy-digital:
Hi @hansfbaier, this is indeed 8192 and common between all SDRAM, sorry for also not noticing it when merging the W8925G6KH6 support.

PR #286, opened 2021-11-24
litedram_gen fixes for ECP5

mkj:
With these three changes I am able to use litedram generated verilog for a standalone build of Microwatt on an OrangeCrab 85F 0.2.

There is one remaining issue I need to still resolve,  https://github.com/enjoy-digital/litedram/commit/e52ece0b8a73eda98c1e2a259000db54703497a8  blocks userport access while memtest runs from `sdram_init` - I can't see how our  standalone build (from same `sdram.c` source) is any different, but for now I've patched it to set `ddrctrl_init_done_write(1)` pre-memtest.

### DIFF
--- a/litedram/gen.py
+++ b/litedram/gen.py
@@ -317,14 +317,19 @@ class LiteDRAMECP5DDRPHYCRG(Module):
         self.sync.por += If(~por_done, por_count.eq(por_count - 1))
 
         # PLL.
+        sys2x_clk_ecsout = Signal()
         self.submodules.pll = pll = ECP5PLL()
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @mkj, this is merged. When adding https://github.com/enjoy-digital/litedram/commit/e52ece0b8a73eda98c1e2a259000db54703497a8 I indeed maybe forgot this specific cas: When the CPU is not directly integrated in the standalone core, the memory is tested directly from the user port so preventing access from it in this case is problematic. We could always force `user_enable` to `1` in this cases if this is easier for you.

PR #292, opened 2022-01-13
litedram_gen: Don't block user port with no CPU

mkj:
Change e52ece0b8a7 blocks access to user ports prior to dram init.
As noted in https://github.com/enjoy-digital/litedram/pull/286#issuecomment-981362385 this is a problem for cpu-less designs, since the initial memtest requires the user port.

Instead enable the user port unconditionally in that case.

### DIFF
--- a/litedram/gen.py
+++ b/litedram/gen.py
@@ -655,7 +655,13 @@ class LiteDRAMCore(SoCCore):
 
         # User ports -------------------------------------------------------------------------------
         user_enable = Signal()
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @mkj, this is indeed an issue with CPU-less controllers. But even in this case, blocking the other ports could be necessary so the fix was probably not flexible enough. This is fixed a bit differently with https://github.com/enjoy-digital/litedram/commit/62abf9ce0c93d4f2a5588f89716d680138cb70db that introduces a `block_until_ready` parameter on ports that can use set to `False` for the port used by the CPU for memtest in the CPU-less case.

PR #300, opened 2022-03-23
Allow for variable DQ/DQS ratio

RRozak:
This PR adds the possibility for variable DQ/DQS ratio. I saw that there was handling of 4x, but this approach is more general and I think it is simpler.


enjoy-digital:
Thanks @RRozak, ok please update the PR when you think it's ready.

RRozak:
I think it is ready. It also requires a change in litex: https://github.com/enjoy-digital/litex/pull/1258

enjoy-digital (approved):
Thanks @RRozak, this looks good and can be merged.

PR #302, opened 2022-03-28
dfi: add possibility to have an external dfi injector

acomodi:
This PR adds another DFI Interface that can be used to inject commands from logic outside of the LiteDRAM controller.

Moreover, it adds some description of the dfi CSRs.


enjoy-digital (approved):
Thanks @acomodi, this looks good and documenting the CSR was also something I wanted to do.

PR #313, opened 2022-10-30
Add MT46H128M16 and change bankmaschine to not use A10 for col addresses.

cklarhorst:
For larger memories, the col encoding needs to skip `A10` because that one is reserved for auto-precharge.
The remaining col encoding bits are then mapped to `A11` and higher.

I think that newer RAM generations are defining additional reserved address bits, e.g. for DDR4 `A12` is used for burst-chop. But I ignored those higher bits, as we don't have any module with that many col bits used.

### DIFF
--- a/litedram/core/bankmachine.py
+++ b/litedram/core/bankmachine.py
@@ -34,7 +34,10 @@ class _AddressSlicer:
 
     def col(self, address):
         split = self.colbits - self.address_align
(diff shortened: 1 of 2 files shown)


cklarhorst:
(Thanks to the unit tests, I found an off by one error in my if)

enjoy-digital (approved):
Thanks @cklarhorst, this is merged. Sorry for the delay.

PR #314, opened 2022-11-14
Fix DFITimingsChecker for DDR4 simulation

michalsieron:
In case of DDR4 tRFC and tREFI timings are actually dictionaries with timings specific for the chosen refresh mode.

Right now, it is impossible to simulate DDR4, because an exception happens in `DFITimingsChecker.prepare_timings` method when indexing `val` variable.

This is due to the fact, that in `DFITimingsChecker.__init__` we request timing values from the module by name, ignoring the fact that some of them (tRFC and tREFI) need to be first accessed using the chosen refresh mode.

This commit fixes this error, by properly using `key` parameter when calling `SDRAMModule.get` method to get only required timing.

If one were to fix it in `DFITimingsChecker.prepare_timings` method, it would require duplicating logic from `SDRAMModule.get` so this is a simpler and cleaner solution.

### DIFF
--- a/litedram/phy/model.py
+++ b/litedram/phy/model.py
@@ -199,17 +199,12 @@ class DFITimingsChecker(Module):
         return self.ns_to_ps(max(c, t))
 
     def prepare_timings(self, timings, refresh_mode, memtype):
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @michalsieron, that's merged.

PR #319, opened 2023-01-04
frontend/bist: properly signal finished writes

michalsieron:
Without it, software was resetting the generator too early and wrong data was being written to the RAM.


enjoy-digital:
Thanks @michalsieron, this is merged.

PR #322, opened 2023-01-11
Make tests safe to run in parallel

michalsieron:
For example using pytest-parallel you can greatly reduce time it takes to run all tests.

```
$ pytest --workers auto test
```


enjoy-digital:
Thanks @michalsieron, sorry for the delay.

PR #326, opened 2023-02-08
phy/gw2ddrphy: supressing warnings about unconnected and bit length.

trabucayre:
When using `gw2ddrphy` **Gowin** tool produces bunch of warnings like:

```
WARN  (EX3073) : Port 'Q1' remains unconnected for this instance("/somewhere/build/sipeed_tang_primer_20k/gateware/sipeed_tang_primer_20k.v":10334)
[...]
WARN  (EX3670) : Actual bit length 1 differs from formal bit length 8 for port 'WSTEP'("/somewhere/build/sipeed_tang_primer_20k/gateware/sipeed_tang_primer_20k.v":10364)
```

It's also true for simulation using **verilator**

This PR add an class called `Open` to handle unconnected output ports, and uses `Constant` to specify input ports size.

### DIFF
--- a/litedram/phy/gw2ddrphy.py
+++ b/litedram/phy/gw2ddrphy.py
@@ -25,6 +25,8 @@ from litex.soc.interconnect.csr import *
 from litedram.common import *
 from litedram.phy.dfi import *
 
+class Open(Signal): pass
+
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre!

PR #335, opened 2023-04-24
Fix AxSIZE

### LINKED ISSUE #334: AxSIZE mismatch?
When generating a ram controller with an Axi interface, the awsize and arsize ports are 4 bits wide 
```verilog
    input  wire    [3:0] user_port_axi_0_awsize,
```
```verilog
    input  wire    [3:0] user_port_axi_0_arsize,
```
but according to section A4.1.1, those fields should be 3 bits wide. This also seems to be how they are used in the generated verilog
```verilog
wire    [2:0] main_aw_payload_size;
assign main_aw_payload_size = user_port_axi_0_awsize;
```

Is this intentional?

TheZoq2:
Closes #334 

As far as I can tell, the axi port generation also assumes size 3, so hopefully this is the right fix

### DIFF
--- a/litedram/gen.py
+++ b/litedram/gen.py
@@ -224,7 +224,7 @@ def get_axi_user_port_ios(_id, aw, dw, iw):
             Subsignal("awaddr",  Pins(aw)),
             Subsignal("awburst", Pins(2)),
             Subsignal("awlen",   Pins(8)),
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @TheZoq2, this looks correct and also related to https://github.com/enjoy-digital/litex/pull/2222 that has also been merged.

PR #351, opened 2023-11-09
phy/gw5ddrphy: introducing GW5A DDR phy

trabucayre:
It's a simple gw2ddrphy copy with fix for some primitives


enjoy-digital:
Thanks @trabucayre!

PR #352, opened 2024-01-11
litedram/phy/lpddr*: fix use of invalid escape sequence

maribu:
This fixes multiple instances of:

litedram/phy/lpddr4/commands.py:209
  litedram-2023.12/litedram/phy/lpddr4/commands.py:209: DeprecationWarning: invalid escape sequence '\d'
    "BA(\d+)": lambda i: self.dfi.bank[i],


enjoy-digital:
Thanks @maribu, this is merged.

PR #354, opened 2024-02-20
Avalon Frontend which actually works

hansfbaier:
I had the following issue with current master in MiSTeX:
The sdram usually initializes correctly, and often ran well for up to ten minutes or so,
but then the Screen (which outputs the framebuffer of a scaler) would go dark,
or show a regular garbage pattern.
As long as the screen output was good, sdram_test would pass on the LiteX BIOS console.
But as soon the screen output was corrupt, sdram_test also would fail on the LiteX BIOS console.
Doing a reset fixed the screen output again, with sdram_test passing again for a couple of minutes,
but then it would become corrupt again. The timeframe this would happen
ranges from instantly after core load to more than ten minutes.
This revision not only has a much cleaner code base,
but also now ran without issues for 6 hours.
Since this PR is made up entirely of your own commits, review should not be difficult.
Is there anything which would speak against merging this into master?

### DIFF
--- a/litedram/frontend/adapter.py
+++ b/litedram/frontend/adapter.py
@@ -81,23 +81,28 @@ class LiteDRAMNativePortDownConverter(Module):
 
         ratio = port_from.data_width//port_to.data_width
         mode  = port_from.mode
-        count = Signal(max=ratio)
+
+        cmd_count = Signal(max=ratio)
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Hi @hansfbaier,

Sorry for the delay, this is merged, thanks for the test/feedback and re-combining the commits.

PR #360, opened 2024-07-18
litedram/address_mapping Add bank_byte_alignment

Dolu1990:
This PR add ControllerSettings.bank_byte_alignment which allow to specifiy how many byte (minimum) there should be between bank address change.

This allow to align it to L2 cache sets size, which improve performances quite a bit on VexiiRiscv :
VexiiRiscv RV64GC with hadware prefetch, 256KB 4way L2, running at 100 Mhz => 
```
without this PR
  Write speed: 197.1MiB/s
   Read speed: 379.4MiB/s 

With this PR and 
            self.add_sdram("sdram",
                phy           = self.ddrphy,
                module        = MT41K256M16(sys_clk_freq, "1:4"),
                l2_cache_size = kwargs.get("l2_size", 8192),
               controller_settings=ControllerSettings(bank_byte_alignment= 256*1024//4) # <-----
            )
  Write speed: 252.3MiB/s
   Read speed: 424.6MiB/s
``` 

When set to 0 (default value) it will not interfer with the previous behaviour of address_mapping     = "ROW_BANK_COL",

### DIFF
--- a/litedram/core/controller.py
+++ b/litedram/core/controller.py
@@ -40,7 +40,13 @@ class ControllerSettings(Settings):
         with_auto_precharge = True,
 
         # Address mapping
-        address_mapping     = "ROW_BANK_COL"):
+        address_mapping     = "ROW_BANK_COL",
+
+        # bank_byte_alignment specify how many bytes should be in between each bank change (minimum).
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @Dolu1990!

PR #361, opened 2024-10-31
frontend/dma.py: Remove left-over rsv_level documentation

david-sawatzke:
Was reworked in a previous commit and rsv_level didn't survive


trabucayre:
Applied. Thanks @david-sawatzke

PR #366, opened 2025-07-02
Added W989D6DBGX6

mmicko:
This adds SDRAM used by Machdyne Kolsch board


trabucayre:
Applied. Thanks @mmicko !

PR #377, opened 2026-04-01
Added last signal to LiteDRAMWriter

Skinbow:
This PR adds a last signal transmission to the `LiteDRAMDMAWriter` to enable writing packets and stopping their transmission.

While extending LiteEth with RDMA with the pull request ["Implementing RoCEv2 passive receiver#198"](https://github.com/enjoy-digital/liteeth/pull/198), I noticed that the writer did not transfer the last signal to the port controller.


enjoy-digital:
Thanks @Skinbow, this is merged.

PR #380, opened 2026-05-12
Add NT5CC256M16DP DDR3 module definition

theox33:
This RAM module is used on the board enclustra_mercury_kx2 which is already documented in the repo `litex-boards` as `litex_boards/targets/enclustra_mercury_kx2.py`.

As there are different version with different RAM modules of the enclustra_mercury_kx2, I wanted to add the NT5CC256M16DP RAM module for better compatibility of the enclustra_mercury_kx2 board.

Just change the `module` attribute in `self.add_sdram(` to NT5CC256M16DP and imports at beginning of the `litex_boards/targets/enclustra_mercury_kx2.py` to change the DDR module to this one.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -1,11 +1,7 @@
 name: PyPI
 
 on:
-  push:
-    tags:
-      - "[0-9][0-9][0-9][0-9].04"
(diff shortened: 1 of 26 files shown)


enjoy-digital:
Thanks @theox33, this is merged.

PR #381, opened 2026-06-20
modules: encode LPDDR4 tFAW nCK floor for MT53E256M16D1

seam0814:
## Summary

The MT53E256M16D1 LPDDR4 module encodes its `tFAW` and `tRFC` as bare
numbers in `litedram/modules.py`:

```python
"1866": _SpeedgradeTimings(... tRFC=180, tFAW=40, ...)
```

`tFAW=40` is interpreted as "40 ns, no nCK constraint". JEDEC JESD209-4
specifies LPDDR4 8-bank `tFAW = max(32 nCK, 40 ns)`, so at controller
clocks low enough that 40 ns rounds to fewer than 32 nCK the resolved
cycle count under-constrains the four-activate window and can let too
many `ACT`s into the rolling window. This is the same class of bug
fixed for the DDR4 RDIMM x4 modules in 183011d (HMA82GR7DJR4N /
M393A2K40DB3).

## Change

```diff
- "1866": _SpeedgradeTimings(... tRFC=180,        tFAW=40,      ...)
+ "1866": _SpeedgradeTimings(... tRFC=(None, 180), tFAW=(32, 40), ...)
```

- `tFAW=(32, 40)` — encodes the JEDEC 32 nCK floor.
- `tRFC=(None, 180)` — equivalent value, switched to the explicit
  tuple form for consistency with the surrounding tuple-form timings
  and with the H5TC4G63CFR fix in 909fe8b. No behaviour change for
  `tRFC`.

## Test

Adds `test_mt53e256m16d1_tfaw_uses_ck_minimum`, mirroring the
existing `test_ddr4_rdimm_tfaw_uses_ck_minimum` and
`test_h5tc4g63cfr_geometry_and_trfc` style:

```python
def test_mt53e256m16d1_tfaw_uses_ck_minimum(self):
    cls = litedram.modules.MT53E256M16D1
    module = cls(clk_freq=100e6, rate="1:4")
    self.assertEqual(cls.speedgrade_timings["1866"].tFAW, (32, 40))
    self.assertEqual(module.timing_settings.tFAW, 8)
```

`python -m unittest test.test_modules` runs all 17 tests cleanly with
the change applied.

PR #382, opened 2026-06-20
modules: bump H5TC4G63CFR tRRD nCK floor to 6 for x16 organisation

seam0814:
## Summary

H5TC4G63CFR was re-modelled as the SK hynix DDR3L 4Gb x16 device in
909fe8b ("modules: fix SK hynix DDR3L geometry"), but its `tRRD`
encoding was left at the x4/x8 nCK floor:

```python
technology_timings = _TechnologyTimings(... tRRD=(4, 7.5) ...)
```

JEDEC DDR3/DDR3L specifies the four-bank `tRRD` nCK floor as **6 nCK
for x16 organisation** (vs 4 nCK for x4 and x8). The sibling x16 SK
hynix part added in a59772c — H5TQ4G63CFR — already uses the correct
value:

```python
technology_timings = _TechnologyTimings(... tRRD=(6, 7.5) ...)
```

Both parts are 4Gb x16 DDR3 from SK hynix in the same H5T family; the
only intended difference is supply voltage (DDR3L vs DDR3), which
does not relax the `tRRD` nCK floor.

## Change

```diff
-    technology_timings = _TechnologyTimings(tREFI=64e6/8192, tWTR=(4, 7.5), tCCD=(4, None), tRRD=(4, 7.5), tZQCS=(64, 80))
+    technology_timings = _TechnologyTimings(tREFI=64e6/8192, tWTR=(4, 7.5), tCCD=(4, None), tRRD=(6, 7.5), tZQCS=(64, 80))
```

Plus a brief inline comment explaining the x16 floor and the
H5TQ4G63CFR cross-reference.

## Test

Extends the existing `test_h5tc4g63cfr_geometry_and_trfc` (added in
909fe8b) with one extra assertion so the regression is locked in:

```python
self.assertEqual(cls.technology_timings.tRRD, (6, 7.5))
```

`python -m unittest test.test_modules` runs all tests cleanly with the
change applied.

## Why this is a separate PR from #381

PR #381 fixes the analogous LPDDR4 `tFAW` 32 nCK floor on
MT53E256M16D1. This PR is unrelated except in flavour — it touches a
different module, a different timing parameter, and a different DRAM
generation, and is cleanly reviewable on its own.

### DIFF
--- a/litedram/modules.py
+++ b/litedram/modules.py
@@ -907,7 +907,9 @@ class H5TC4G63CFR(DDR3Module):
     nrows  = 32768
     ncols  = 1024
     # timings
-    technology_timings = _TechnologyTimings(tREFI=64e6/8192, tWTR=(4, 7.5), tCCD=(4, None), tRRD=(4, 7.5), tZQCS=(64, 80))
+    # JEDEC DDR3 x16 organisation requires the 6 nCK tRRD floor (vs 4 nCK
+    # for x4/x8). This matches the sibling x16 H5TQ4G63CFR definition.
+    technology_timings = _TechnologyTimings(tREFI=64e6/8192, tWTR=(4, 7.5), tCCD=(4, None), tRRD=(6, 7.5), tZQCS=(64, 80))
     speedgrade_timings = {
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @seam0814, this is merged.

PR #384, opened 2026-06-24
modules: bump H5TC4G63CFR -800 tFAW to JEDEC x16 50 ns floor

seam0814:
## Summary

H5TC4G63CFR was re-modelled as the SK hynix DDR3L 4Gb x16 device in
909fe8b (\"modules: fix SK hynix DDR3L geometry\"), and #382 corrected
its \`tRRD\` nCK floor to the x16 value. The \`-800\` \`tFAW\` ns floor is
still encoded with the x4/x8 value:

\`\`\`python
\"800\":  _SpeedgradeTimings(... tFAW=(None, 40), ...)
\`\`\`

JEDEC DDR3/DDR3L specifies the four-bank window \`tFAW\` ns floor as
**50 ns for DDR3-800 x16 organisation** (vs 40 ns for x4 and x8).
Every other x16 DDR3-800 module in \`litedram/modules.py\` already
encodes the 50 ns floor — MT41J128M16, MT41J256M16, MT41K64M16,
K4B2G1646F — so H5TC4G63CFR is the lone outlier.

## Change

\`\`\`diff
-        \"800\":  _SpeedgradeTimings(tRP=15, tRCD=15, tWR=15, tRFC=(None, 260), tFAW=(None, 40), tRAS=37.5),
+        \"800\":  _SpeedgradeTimings(tRP=15, tRCD=15, tWR=15, tRFC=(None, 260), tFAW=(None, 50), tRAS=37.5),
\`\`\`

Plus a brief inline comment recording the JEDEC x16 floor for future
maintainers.

## Test

Extends the existing \`test_h5tc4g63cfr_geometry_and_trfc\` (added in
909fe8b, further covered in #382) with one extra assertion locking
the corrected value:

\`\`\`python
self.assertEqual(cls.speedgrade_timings[\"800\"].tFAW, (None, 50))
\`\`\`

\`python -m unittest test.test_modules\` runs all 16 tests cleanly
with the change applied.

## Why this is separate from #382

#382 bumps the **technology-scope** \`tRRD\` nCK floor; this PR bumps
a **speedgrade-scope** \`tFAW\` ns floor in the \`-800\` entry. Each is
a one-line edit driven by the same x16 JEDEC rule on a different
parameter, and they are cleanly reviewable independently.

PR #394, opened 2026-07-05
test: lock in DDR3-800 x16 tFAW >= 50 ns floor across the family (regression)

seam0814:
## Summary

#384 corrected a single x16 SK hynix part (H5TC4G63CFR) whose
DDR3-800 \`tFAW\` was still encoded with the x4/x8 40 ns floor
instead of the JEDEC x16 50 ns floor. The rest of the DDR3 x16
devices in \`litedram/modules.py\` already encode 50 ns at that
speedgrade — the fix was to bring H5TC4G63CFR into line with them,
not to change the floor itself.

This test-only PR locks that invariant in going forward: every
DDR3 x16 device that exposes a \`\"800\"\` speedgrade must have
\`tFAW\` ns >= 50 (JEDEC DDR3-800 x16 minimum, vs 40 ns for x4/x8).
Any future module addition or accidental \`(None, 40)\` regression on
an x16 part will now fail the test loudly rather than silently
under-constrain the four-ACT window at DDR3-800.

## What's added

One new test, \`test_ddr3_x16_tfaw_800_ns_minimum\`, in the same
style as the existing \`test_ddr3_x16_trrd_uses_ck_minimum\`:

\`\`\`python
def test_ddr3_x16_tfaw_800_ns_minimum(self):
    names = [
        \"MT41K64M16\",
        \"MT41J128M16\", \"MT41K128M16\",
        \"MT41J256M16\", \"MT41K256M16\",
        \"K4B2G1646F\",
        \"H5TC4G63CFR\",
        \"IS43TR16512B\",
    ]
    for name in names:
        cls = getattr(litedram.modules, name)
        with self.subTest(module=name):
            self.assertIn(\"800\", cls.speedgrade_timings)
            tfaw = cls.speedgrade_timings[\"800\"].tFAW
            self.assertGreaterEqual(tfaw[1], 50, ...)
\`\`\`

## Why this is worth its own PR

Pure test-only — **no behaviour change**. The invariant it locks in
was established by an already-merged fix (#384) and is currently
satisfied by every listed part; the test simply codifies the rule so
it stays true. Kept as its own PR (rather than folded into a module
fix) so the diff is unambiguously additive and can be reviewed in a
few seconds.

## Test run

\`python -m unittest test.test_modules\` runs cleanly on \`master\`
with the new test added — all listed modules already meet the floor.

### DIFF
--- a/litedram/modules.py
+++ b/litedram/modules.py
@@ -945,7 +945,8 @@ class IS43TR16128B(DDR3Module):
     nrows  = 16384
     ncols  = 1024
     # timings
-    technology_timings = _TechnologyTimings(tREFI=64e6/8192, tWTR=(4, 7.5), tCCD=(4, None), tRRD=(4, 6), tZQCS=(64, 80))
+    # JEDEC DDR3 x16 organisation requires the 6 nCK tRRD floor (vs 4 nCK for x4/x8).
+    technology_timings = _TechnologyTimings(tREFI=64e6/8192, tWTR=(4, 7.5), tCCD=(4, None), tRRD=(6, 6), tZQCS=(64, 80))
     speedgrade_timings = {
         "1600": _SpeedgradeTimings(tRP=13.75, tRCD=13.75, tWR=15, tRFC=(None, 160), tFAW=(None, 40), tRAS=35),
     }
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @seam0814.

PR #393, opened 2026-07-05
modules: encode DDR4 x16 tFAW nCK floor for MTA4ATF51264HZ (extends #392)

seam0814:
## Summary

The MTA4ATF51264HZ DDR4 SODIMM in `litedram/modules.py` is
characterised with the DDR4 x16 per-device organisation
(`ngroups=2, ngroupbanks=4`) — the same shape used for MT40A256M16
and MT40A512M16 — but its \`tFAW\` at DDR4-2133 still uses the x4/x8
nCK floor:

\`\`\`python
class MTA4ATF51264HZ(DDR4Module):
    ngroupbanks = 4
    ngroups     = 2                                      # x16 shape
    ...
    \"2133\": _SpeedgradeTimings(... tFAW=(20, 25), ...)   # x4/x8 floor
\`\`\`

JEDEC JESD79-4 fixes the four-bank window \`tFAW\` **nCK floor at
28 nCK for x16 organisation** (vs 20 nCK for x4 and x8) — the same
class of bug fixed for MT40A512M16 in #392. At controller clocks
low enough that the ns side rounds to fewer than 28 nCK the resolved
cycle count under-constrains the four-ACT window.

## Change

Bump only the nCK element of \`tFAW\` from 20 to 28, matching the
convention established by #381, #384 and #392 (touch nCK only, leave
datasheet-derived ns alone).

\`\`\`diff
-        \"2133\": _SpeedgradeTimings(... tFAW=(20, 25), ...)
+        # ngroups=2 characterises this SODIMM as built from DDR4 x16 devices.
+        # JEDEC DDR4 x16 organisation requires the 28 nCK tFAW floor (vs 20 nCK for x4/x8).
+        \"2133\": _SpeedgradeTimings(... tFAW=(28, 25), ...)
\`\`\`

## Test

Adds \`test_mta4atf51264hz_tfaw_uses_x16_ck_minimum\`, mirroring the
style of the MT40A512M16 test added in #392:

\`\`\`python
def test_mta4atf51264hz_tfaw_uses_x16_ck_minimum(self):
    cls = litedram.modules.MTA4ATF51264HZ
    module = cls(clk_freq=100e6, rate=\"1:4\")
    self.assertEqual(cls.speedgrade_timings[\"2133\"].tFAW[0], 28)
\`\`\`

\`python -m unittest test.test_modules\` runs cleanly with the
change applied.

## Why this is a separate PR from #392

#392 targets the MT40A512M16 bare-chip DDR4 x16 outlier. This PR
[...]

PR #392, opened 2026-07-05
modules: encode DDR4 x16 tFAW nCK floor for MT40A512M16

seam0814:
## Summary

The MT40A512M16 DDR4 x16 module in `litedram/modules.py` encodes its
`tFAW` at DDR4-2400 with the x4/x8 nCK floor:

```python
class MT40A512M16(DDR4Module):
    ngroupbanks = 4
    ngroups     = 2                                      # x16 shape
    ...
    "2400": _SpeedgradeTimings(... tFAW=(20, 25), ...)   # x4/x8 floor
```

JEDEC JESD79-4 fixes the four-bank window `tFAW` **nCK floor at
28 nCK for x16 organisation** (vs 20 nCK for x4 and x8) — the same
class of bug as #381 (LPDDR4 tFAW nCK) and #384 (DDR3 x16 tFAW ns).
At controller clocks low enough that the ns side rounds to fewer
than 28 nCK the resolved cycle count under-constrains the four-ACT
window.

The sibling `MT40A256M16`, which is the same Micron DDR4 x16 family
(`ngroups=2, ngroupbanks=4`) at DDR4-2400, already encodes the
correct floor:

```python
class MT40A256M16(DDR4Module):
    ...
    "2400": _SpeedgradeTimings(... tFAW=(28, 35), ...)   # x16 floor
```

so MT40A512M16 is the lone outlier among the two Micron DDR4 x16
parts in the module list.

## Change

Bump only the nCK element of `tFAW` from 20 to 28. The ns element is
left at the per-device value (25) matching the convention established
by #381/#384 (touch nCK only, leave datasheet-derived ns alone). If a
future PR wants to align the ns side with the sibling MT40A256M16
(35 ns), that can be done independently under its own review.

```diff
-        \"2400\": _SpeedgradeTimings(... tFAW=(20, 25), ...)
+        # JEDEC DDR4 x16 organisation requires the 28 nCK tFAW floor (vs 20 nCK for x4/x8).
+        \"2400\": _SpeedgradeTimings(... tFAW=(28, 25), ...)
```

## Test

Adds `test_mt40a512m16_tfaw_uses_x16_ck_minimum`, mirroring the style
of `test_mt53e256m16d1_tfaw_uses_ck_minimum` (LPDDR4, #381) and
`test_ddr4_rdimm_tfaw_uses_ck_minimum` (DDR4 RDIMM):

```python
def test_mt40a512m16_tfaw_uses_x16_ck_minimum(self):
[...]

PR #395, opened 2026-07-12
test: lock in DDR4 x16 tFAW >= 28 nCK floor across the family (regression)

seam0814:
## Summary

#392 and #393 corrected the two DDR4 x16 outliers in this repo whose
\`tFAW\` was still encoded with the x4/x8 20 nCK floor instead of the
JEDEC (JESD79-4) x16 28 nCK floor: **MT40A512M16** (bare chip, #392)
and **MTA4ATF51264HZ** (SODIMM, #393). \`MT40A256M16\` already carried
the correct 28 nCK floor.

This test-only PR locks the invariant in going forward: every DDR4
module whose per-device organisation is characterised as x16 in this
repo (\`ngroups == 2, ngroupbanks == 4\`) must have \`tFAW\` nCK >= 28
at every speedgrade it exposes. Any future addition or accidental
\`(20, ...)\` regression on an x16 part will now fail loudly rather
than silently under-constrain the four-ACT window.

## What's added

One new test, \`test_ddr4_x16_tfaw_uses_ck_minimum\`, in the same
style as \`test_ddr3_x16_tfaw_800_ns_minimum\` added in #394:

\`\`\`python
def test_ddr4_x16_tfaw_uses_ck_minimum(self):
    names = [
        \"EDY4016A\",
        \"MT40A256M16\",
        \"MT40A512M16\",
        \"MTA4ATF51264HZ\",
    ]
    for name in names:
        cls = getattr(litedram.modules, name)
        with self.subTest(module=name):
            self.assertEqual(cls.ngroups, 2, ...)   # confirm x16 shape
            for sg_name, sg in cls.speedgrade_timings.items():
                if sg_name == \"default\":
                    continue
                self.assertGreaterEqual(sg.tFAW[0], 28, ...)
\`\`\`

The test iterates every speedgrade of every listed class (instead of
hard-coding \"2400\" or \"2133\"), so a future speedgrade addition
can't accidentally slip through with a 20 nCK floor.

## Modules covered

All four DDR4 modules whose per-device organisation is x16 in this
repo (\`ngroups == 2\`):

- \`EDY4016A\` (Elpida DDR4)
- \`MT40A256M16\` (Micron DDR4)
- \`MT40A512M16\` (Micron DDR4, fixed in #392)
- \`MTA4ATF51264HZ\` (Micron DDR4 SODIMM, fixed in #393)

## Why this is worth its own PR

Pure test-only — **no behaviour change**. The invariant it locks in
was established by already-merged fixes (#392 / #393) and is
currently satisfied by every listed part; the test simply codifies
the rule so it stays true. Kept as its own PR (rather than folded
into a module fix) so the diff is unambiguously additive and can be
reviewed in seconds, mirroring the split used in #394.

## Test run

\`python -m unittest test.test_modules\` runs cleanly on \`master\`
with the new test added — all listed modules already meet the floor.

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -86,6 +86,40 @@ class TestSDRAMModules(unittest.TestCase):
         module = cls(clk_freq=100e6, rate="1:4")
         self.assertEqual(cls.speedgrade_timings["2133"].tFAW[0], 28)
 
+    def test_ddr4_x16_tfaw_uses_ck_minimum(self):
+        # JEDEC DDR4 (JESD79-4) fixes the tFAW nCK floor at 28 for x16
+        # organisation (vs 20 nCK for x4/x8). PRs #392 and #393 corrected
+        # the two DDR4 x16 outliers in this repo (MT40A512M16 bare-chip
+        # and MTA4ATF51264HZ SODIMM); this regression test locks the
+        # invariant in across every DDR4 device whose per-device
+        # organisation is x16 (`ngroups == 2`, `ngroupbanks == 4`), so
+        # any future addition or accidental (20, ...) regression fails
+        # loudly at test time.
+        names = [
+            "EDY4016A",
+            "MT40A256M16",
+            "MT40A512M16",
+            "MTA4ATF51264HZ",
+        ]
+        for name in names:
+            cls = getattr(litedram.modules, name)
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @seam0814!

PR #396, opened 2026-07-21
Add DDR3 PHY for Lattice Nexus FPGAs

gatecat:
Tested with Radiant 2025.2 on the Certus-NX Versa board (https://github.com/litex-hub/litex-boards/pull/757), nextpnr support hopefully coming very soon too.

This is mostly cribbed from the ECP5 PHY with changes to latencies where necessary. The part I'm least sure about is the half-system-clock (i.e. 1 memory clock) latency difference for writes. Currently this is dealt with by adding one to CWL but I'm not sure if a write bitslip on DQ/DQS (like 7-series) is overall a more robust solution to this.

In the future this could also be extended to support 8:1 mode as the Nexus devices have

### DIFF
--- a/litedram/phy/__init__.py
+++ b/litedram/phy/__init__.py
@@ -5,6 +5,8 @@ from litedram.phy.s7ddrphy import V7DDRPHY, K7DDRPHY, A7DDRPHY
 from litedram.phy.usddrphy import USDDRPHY, USPDDRPHY
 
 from litedram.phy.ecp5ddrphy import ECP5DDRPHY, ECP5DDRPHYInit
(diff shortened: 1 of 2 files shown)


gatecat:
... and with https://github.com/YosysHQ/nextpnr/commit/c2cbdc7efac2b40e276c1c387106f9304287c3b7 the nextpnr support should be in place (edge clock routing might not be 100% reliable but I'll fix that soon too)

enjoy-digital:
Awesome, thanks @gatecat! This this is a first good step that we can merge (I just applied cosmetic alignments to it). I'll try to suggest a follow up PR for the cleanup/latency improvement to have a proper solution.

Issue #299, opened 2022-03-21
ULX4M DM signal is not connected to DQS group


goran-mahovlic:
Hi, 

I am trying to get DDR3 running on ULX4M, I am currently using gsd_butterstick.py with all changes needed for ULX4M, but I did not know that DM also needs to be in the same DQS group so I have changed this 

https://github.com/enjoy-digital/litedram/blob/master/litedram/phy/ecp5ddrphy.py#L339

to use

ODDRX2F

But I get memtest fail, so maybe you know out of your hear should this be ok quick fix, and how could I do more debugging to get more info what is wrong if it is not because DM signals.

![memtest_KO](https://user-images.githubusercontent.com/4248736/159245575-dccdf397-c518-49f6-bfc8-3beb4302bd9f.png)

enjoy-digital:
Hi @goran-mahovlic,

the BIOS `mem_test `is only doing full access to the DRAM (no use of DM), so in a first time, can you comment this: https://github.com/enjoy-digital/litedram/blob/master/litedram/phy/ecp5ddrphy.py#L339-L346 and replace it with:
`self.comb += pads.dm[i].eq(0)`

This will avoid latency related issue on DM and will help knowing if the issue is related to DM or something else.

goran-mahovlic:
> Hi @goran-mahovlic,
> 
> the BIOS `mem_test `is only doing full access to the DRAM (no use of DM), so in a first time, can you comment this: https://github.com/enjoy-digital/litedram/blob/master/litedram/phy/ecp5ddrphy.py#L339-L346 and replace it with: `self.comb += pads.dm[i].eq(0)`
> 
> This will avoid latency related issue on DM and will help knowing if the issue is related to DM or something else.

Thank you, that works!

![genius](https://user-images.githubusercontent.com/4248736/159274191-faa06fa2-bf31-4170-bc90-4ca71c071c35.png)

PR #401, opened 2026-08-01
phy/ecp5ddrphy.py: added argument and logic to skip dm management when dm pads are not in the same group than rest of DRAM pads

trabucayre:
Simple hack based on https://github.com/enjoy-digital/litedram/issues/299
When dm can't be used signals are fixed to 0.
The target must uses an l2_cache.
Tested with radiona_ulx4m_ld to boot linux

### DIFF
--- a/litedram/phy/ecp5ddrphy.py
+++ b/litedram/phy/ecp5ddrphy.py
@@ -119,7 +119,10 @@ class ECP5DDRPHY(Module, AutoCSR):
         cwl          = None,
         cmd_delay    = 0,
         clk_polarity = 0,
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Added follow-up commit e1230f4 with the suggested simplification: renamed dm_skip to the positive with_dm convention, propagated it to PhySettings, reduced the warning to one concise full-width-write comment, and added a regression test checking both the reported capability and ODDRX2DQA removal. Validation: pytest -q test/test_ddr3_phy_settings.py passes (3 tests, 36 subtests). The refreshed PR CI is now running.

